PR #8, opened 2017-09-27
wildcard intents in stories

ricwo:
**Proposed changes**:
- wildcard intents beginning with `* NOT`
- training data are generated for all other intents defined in the domain

example:
```
* NOT _gender[gender=answer]
```

- [] made PR ready for code review
- [ ] updated the changelog


tmbo:
I think we'll not include this for now.

PR #10, opened 2017-10-06
fixed tracker restart history issues

tmbo:
**Proposed changes**:
- previously for every loading of a persisted dialogue more `action_listen` events got appended after a restart leading to a messed up feature history - fixed that

- [x] updated the changelog


ricwo:
`tracker.events` can now be `None`, which isn't yet accounted for in other places. But otherwise this fixes it

We can either write a an `ActionExecuted(ActionListen())` in the tracker as a first event after a reset (ignoring all others), or  catch `events is None` everywhere else

PR #28, opened 2017-10-14
Enable loading of inherited Agent classes

pawni:
Currently the Agent class can only load instances of itself and not from inherited classes. The `load` function should return `cls(..)` and not `Agent(..)` to enable this.

**Proposed changes**:
- Change return of `Agent.load` from `Agent(..)` to `cls(..)`

- [ ] updated the changelog


amn41:
looks good to me! once tests pass feel free to merge

PR #31, opened 2017-10-16
 add _save_tracker in 'else' block for the event_restart and event_action_listen

ahand520:
refer  https://github.com/RasaHQ/rasa_core/issues/30

 for event restart and event action_listen

**Proposed changes**:
- 

- [ ] updated the changelog

### DIFF
--- a/rasa_core/processor.py
+++ b/rasa_core/processor.py
@@ -121,13 +121,15 @@ class MessageProcessor(object):
         self._log_action_on_tracker(tracker, executed_action, events)
         if self._should_predict_another_action(executed_action, events):
             action = self._get_next_action(tracker)
+            # save tracker state to continue conversation from this state    
+            self._save_tracker(tracker)
             return {"next_action": action.name(),
                     "tracker": tracker.current_state()}
         else:
+            self._save_tracker(tracker)
             return {"next_action": None,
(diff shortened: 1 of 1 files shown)


rasa_core/processor.py
```diff
@@ -121,14 +121,13 @@ def continue_message_handling(self, sender_id, executed_action, events):
...
             action = self._get_next_action(tracker)
-            # save tracker state to continue conversation from this state
-            self._save_tracker(tracker)
             return {"next_action": action.name(),
                     "tracker": tracker.current_state()}
         else:
             return {"next_action": None,
                     "tracker": tracker.current_state()}
-
+        # save tracker state to continue conversation from this state
```
tmbo: Since this after the `return`s of all the branches, this will never be executed. I don't see why you moved it there.

ahand520: Oh, It's my miss, I would like to add _save_tracker in 'else' block for the event_restart and event_action_listen, I have a new commit, Sorry.

tmbo:
Thanks for providing this contribution, but I don't think that issue still exists. I think that should be fixed by https://github.com/RasaHQ/rasa_core/pull/43/commits/31f8265d7e472c13e002f7735b644c28318f5fd7

PR #32, opened 2017-10-16
Modify examples: 1. path problems, 2. each example becomes independent

fakesecret:
**Issues**: 
- we can only run examples when "pwd" is ...PATH.../rasa_core/; if not, almost all python scripts within *rasa_core/examples* raise errors because relative paths are used and they cannot be found (for example, *examples/concerts/run.py*).
- some example folders are not independent, such as "fake_user", because (1) some data files are not located in the example folder; (2) python modules within an example are imported from rasa-core directly

**Proposed changes**:
- os.path.abspath((os.path.dirname(__file__))) is used to always correctly locate required files for each example
- Some required files which were not located in the example folder, are now copied into the example folder, such as *fake_user/data/babi_task5_fu_rasa_fewer_actions.md*
- some modules become project modules rather than rasa-core modules, ex: **from examples.concerts.policy import ConcertPolicy** to **from concerts.policy import ConcertPolicy**

As a result, users can make a copy of any example folder, to learn or to begin a project with.

- [ ] updated the changelog


tmbo:
This is a very valuable insight and fix. Unfortunately, we were already in the midst of changing the examples. We cleaned them up and moved them around and updated the documentation, so I am afraid we can not directly apply this PR anymore.

Do you mind testing if the current setup solves that problem? Otherwise it would be great if you could apply the changes to the current master.

tmbo:
Although, I think this is very valuable, I think there are to many conflicts due to the large changes in the examples. I will close this one, but I'd be very happy if you take a look at the examples again.

PR #43, opened 2017-10-21
Minor fixes

tmbo:
**Proposed changes**:
- a  couple of minor style fixes
- fix http bug (wrong state if action listen is not confirmed by api user) #42 
- small dropout in keras model to make it overfit a bit less

- [x] updated the changelog

### DIFF
--- a/rasa_core/policies/keras_policy.py
+++ b/rasa_core/policies/keras_policy.py
@@ -70,7 +70,7 @@ class KerasPolicy(Policy):
         # Build Model
         model = Sequential()
         model.add(Masking(-1, batch_input_shape=batch_shape))
-        model.add(LSTM(n_hidden, batch_input_shape=batch_shape))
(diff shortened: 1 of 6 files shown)


rasa_core/policies/keras_policy.py
```diff
@@ -70,7 +70,7 @@ def _build_model(self, num_features, num_actions, max_history_len):
         # Build Model
         model = Sequential()
         model.add(Masking(-1, batch_input_shape=batch_shape))
-        model.add(LSTM(n_hidden, batch_input_shape=batch_shape))
+        model.add(LSTM(n_hidden, batch_input_shape=batch_shape, dropout=0.2))
```
amn41: please double check whether this means "keep 20%" or "drop 20%" - varies between DL frameworks

tmbo: that is the fraction to drop

PR #45, opened 2017-10-21
'run' script for easy model running

tmbo:
**Proposed changes**:
- small script to load a trained dialogue model with a nlu model to run a bot listening on an input channel (currently cmdline or facebook)

- [x] updated the changelog


tmbo:
will be the basis for a rewrite & simplification of parts of the docs (e.g. we can now do a way simpler hello world that doesn't need any python coding other than for custom actions)

PR #44, opened 2017-10-21
allow templates to contain buttons

tmbo:
**Proposed changes**:
- allows to specify buttons & images in utter templates:
```yml
templates:
  utter_greeting:
  - text: "Hey! How are you?"
    buttons:
    - title: "great"
      payload: "great"
    - title: "super sad"
      payload: "super sad"
  utter_cheer_up:
  - text: "Here is something to cheer you up:"
    image: "https://cdn77.eatliver.com/wp-content/uploads/2017/10/trump-frog.jpg"

```
**Status (please check what you already did)**:
- [x] made PR ready for code review
- [x] added some tests for the functionality
- [x] updated the documentation
- [x] updated the changelog

PR #51, opened 2017-10-23
Domain validation

tmbo:
**Proposed changes**:
- validates the domain yaml file and raises an exception if the file has an improper format (prevents strange errors if e.g. the utterance templates are an array instead of a map)

- [x] updated the changelog

### DIFF
--- a/rasa_core/dispatcher.py
+++ b/rasa_core/dispatcher.py
@@ -44,11 +44,12 @@ class Dispatcher(object):
         if message.get("buttons"):
             self.utter_button_message(message.get("text"),
                                       message.get("buttons"))
-        elif message.get("image"):
-            self.utter_message(message.get("text"))
-            self.utter_attachment(message.get("image"))
         else:
             self.utter_message(message.get("text"))
+
+        # if there is an image we handle it separately as an attachment
+        if message.get("image"):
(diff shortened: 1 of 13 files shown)


rasa_core/dispatcher.py
```diff
@@ -37,12 +37,25 @@ def __init__(self, sender, output_channel, domain):
...
+        # type: (Dict[Text, Any]) -> None
         """Send a message to the client."""
 
+        if message.get("buttons"):
+            self.utter_button_message(message.get("text"),
+                                      message.get("buttons"))
+        elif message.get("image"):
+            self.utter_message(message.get("text"))
+            self.utter_attachment(message.get("image"))
+        else:
```
akelad: what if the user specifies both buttons and an image?

tmbo: very good point :+1:

tmbo:
@akelad btw. this PR includes #44 so if you reviewed this one, you can skip the other one

znat:
This breaks something for me: 
When I run a training, `TemplateDomain` looks for`domain_schema.yml` in the current execution path, not the rasa_core path and can't find it. Any idea how to come around that?

tmbo:
@znat that should be fixed in the latest master

PR #52, opened 2017-10-23
Allow to use the train script to start the simple online learning mode on cmd

tmbo:
**Proposed changes**:
- allows to use the `train.py` script to start online learning using the `--online` flag

- [ ] updated the changelog

### DIFF
--- a/rasa_core/train.py
+++ b/rasa_core/train.py
@@ -5,9 +5,12 @@ from __future__ import unicode_literals
 
 import argparse
 import logging
+
 from builtins import str
 
 from rasa_core.agent import Agent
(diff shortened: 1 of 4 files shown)


rasa_core/train.py
```diff
@@ -47,21 +55,37 @@ def create_argument_parser():
             default=20,
             help="number of training samples to put into one training batch")
     parser.add_argument(
+            '--online',
```
akelad: if there's a short option for --nlu and --verbose, should there not also be one for --online? e.g -o ?

tmbo: `-o` is already used for `output` in other scripts of Rasa, so to keep it consistent I'd avoid a short letter

PR #57, opened 2017-10-27
Fix interpreter format doc

drfb:
**Proposed changes**:

Fix documentation for Interpreter's return format. **intent** should be a dictionary containing **name** and **confidence**. In the current documentation, the example **intent** provided is a string that will cause `AttributeError: 'str' object has no attribute 'get'` error.

**Status**:


tmbo:
good point :+1: thanks for fixing that.

we are currently putting a contributors agreement in place to make sure everything is sound from a legal perspective. I'll get back to you at the end of the week

tmbo:
in the meantime - can you please merge the `master` branch into your branch (or rebase on the most recent version)?

drfb:
I messed it up. Sorry. I created a new PR. #64

PR #64, opened 2017-11-02
Fix interpreter doc format

drfb:
**Proposed changes**:

Fix documentation for Interpreter's return format. **intent** should be a dictionary containing **name** and **confidence**. In the current documentation, the example **intent** provided is a string that will cause `AttributeError: 'str' object has no attribute 'get'` error.

**Status**:

### DIFF
--- a/docs/interpreters.rst
+++ b/docs/interpreters.rst
@@ -18,7 +18,10 @@ which has a method ``parse(message)`` which takes a single string argument and r
 
     {
       "text": "show me chinese restaurants",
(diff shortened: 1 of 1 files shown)


tmbo:
Good point :+1: Would you mind signing the contributor agreement so we can get this merged? https://cla-assistant.io/RasaHQ/rasa_core

drfb:
signed

tmbo:
Great, thanks a lot :rocket:

PR #74, opened 2017-11-06
Fix bug in ProbabilisticFeaturizer encoding.

benjamin-work:
**Proposed changes**:
Previously, calling `.encode` with intent features would result in an
error because `.values()` was called when `.(iter)items()` was
required. Furthermore, there was no point in treating intent features
differently.

Additionally, fix docstring and add unit test.


benjamin-work:
Note: It should be considered to use `six.iteritems(active_features)` instead of `active_features.items()`, since the latter does not do the same thing in Python 2 and Python 3.

tmbo:
Looking good, can you please make sure to sign the CLA (https://cla-assistant.io/RasaHQ/rasa_core)  so I can get this merged :+1:

PR #77, opened 2017-11-07
Fb logging

akelad:
**Proposed changes**:
- Add warnings to log if fb_secret/fb_verify is incorrect

- [ ] updated the changelog

### DIFF
--- a/rasa_core/channels/facebook.py
+++ b/rasa_core/channels/facebook.py
@@ -96,7 +96,7 @@ class FacebookInput(HttpInputComponent):
                 signature = request.headers.get("X-Hub-Signature") or ''
                 if not validate_hub_signature(self.fb_secret, request.data,
                                               signature):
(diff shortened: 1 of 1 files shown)


rasa_core/channels/facebook.py
```diff
@@ -88,12 +88,14 @@ def hello():
                 if request.args.get("hub.verify_token") == self.fb_verify:
                     return request.args.get("hub.challenge")
                 else:
+                    logger.debug("Failure, invalid token")
```
tmbo: let's make them `logger.warn` so they always show up and be a little more verbose (keep in mind that not everyone knows that code and what happened when that message gets raised - so the message should explain what is going wrong.)

akelad: good point 👍

tmbo:
@akelad also please make sure to sign https://cla-assistant.io/RasaHQ/rasa_core :D

PR #78, opened 2017-11-07
Entity retrieval

tmbo:
**Proposed changes**:
- adds a method to the tracker to retrieve the values for an entity from the latest message, e.g. `tracker.get_latest_entity_values("number")` will be `["3"]`
- related to #68
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,7 @@ This project adheres to `Semantic Versioning`_ starting with version 0.2.0.
 
 Added
 -----
+- added method to tracker to retrieve the latest entities #68
 
 Changed
 -------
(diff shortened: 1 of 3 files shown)


fakesecret:
I found you'd created a RP to solve my reported issue #68 Point 2. Thank you. I'd like to remind you that your method can retrieve a list of values for a given entity_name, **in the same order with their positions in the latest message**, if one ner is used for rasa-NLU. For example, "What is 2 plus 3?" leads to `tracker.get_latest_entity_values("number") == ["2", "3"]`.
This method provides an easy way to work with entities, similar with the way for slots. However, professional users may want to directly use entities (list of dict) to get more information about positions, ner types, etc.. Personally, I will still use tracker.latest_message.entities.

tmbo:
Right, I mean this is just a shorthand for "give me the email address the user hopefully added to his latest message". For everything else there will always be the access to the complete nlu results via the `tracker.latest_message` object.

PR #82, opened 2017-11-08
issue56: 'sender' to 'sender_id', 'default' by default

fakesecret:
**Proposed changes**:
See https://github.com/RasaHQ/rasa_core/issues/56. For ten python scripts (as well as some in tests folder):
"sender" is renamed as "sender_id";
"DEFAULT_SENDER" is renamed as "DEFAULT_SENDER_ID";
"sender" is set to DEFAULT_SENDER_ID (= "default")

- [x] updated the changelog

### DIFF
--- a/rasa_core/agent.py
+++ b/rasa_core/agent.py
@@ -61,7 +61,7 @@ class Agent(object):
             text_message,  # type: Text
             message_preprocessor=None,  # type: Optional[Callable[[Text], Text]]
             output_channel=None,  # type: Optional[OutputChannel]
-            sender=None  # type: Optional[Text]
+            sender_id="default"  # type: Optional[Text]
     ):
         # type: (...) -> Optional[List[Text]]
         """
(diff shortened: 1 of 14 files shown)


rasa_core/agent.py
```diff
@@ -61,7 +61,7 @@ def handle_message(
             text_message,  # type: Text
             message_preprocessor=None,  # type: Optional[Callable[[Text], Text]]
             output_channel=None,  # type: Optional[OutputChannel]
-            sender=None  # type: Optional[Text]
+            sender_id="default"  # type: Optional[Text]
```
tmbo: How about setting it to `sender_id=DEFAULT_SENDER_ID ` as I'd rather keep the `"default"` only in one place.

For documentation reasons, we can still add the actual value to the documentation of that function

tmbo:
Please add that change to the `Changelog.rst` under changed.

rasa_core/agent.py
fakesecret: I also thought of this before. But I didn't do it because the TrackerStore (and 2 other inherited classes)'s retrieve method don't have default value for sender_id, see below:
```
class TrackerStore(object):
 ......
    def retrieve(self, sender_id):
        raise NotImplementedError()
```
That means if we use default value of sender_id : rasa_core.channels.UserMessage.DEFAULT_SENDER_ID in handle_message method:
```
lst_out_msg = agent.handle_message(input_msg) # sender_id = DEFAULT_SENDER_ID
# users cannot easily know what is DEFAULT_SENDER_ID, unless they go through 2 or 3 scripts
tracker = self.agent.tracker_store.retrieve(sender_id=???) 
```

Do you see my concern for user experience? Ideally, DEFAULT_SENDER_ID should be in a more suitable place. Secondly, many methods' sender_id should have this default value. Tell me what you think so that I can finish the RP

tmbo: Got you. I wouldn't mind if you put the constant somewhere else. But I still would like to use that constant instead of the string. How about adding the default to the tracker store as well?

Ideally there would be a function comment that explains that for the sender parameter :D

tmbo:
Nice job 🚀

PR #98, opened 2017-11-15
Implement new intent entity format

tmbo:
**Proposed changes**:
- do not prepend user intents with `_` anymore in story files
- annotate entities with json objects instead of `[name=Rasa]` it would be `{"name": "Rasa"}` (makes it consistent with the way slots are set)
- use slash notation if you want to set intent / entities from an input channel, e.g. `/greet{"name": "Rasa"}`
- removed deprecated slot setting format (no `- slot["address"]` anymore, just `- slot{"address": "5th Street"}`)

- [x] updated the changelog


amn41:
what do you mean by `use slash notation if you want to set intent / entities from an input channel`  ?

tmbo:
If you, e.g. on the cmdline, want to enter an intent directly instead of a user message, you'd now say `_greet[name=Rasa]` whereas with this implementation it would be `/greet{"name": "Rasa"}`

znat:
Do you by chance have a script to convert story files to the new intent/entities format?

PR #102, opened 2017-11-19
WIP: logging all bot utterances in the tracker

ricwo:
**Proposed changes**:
- unfeaturized `BotUttered` event
- this aims to make writing conversation logs easier, as they can be retrieved from the `Tracker`, independent of output channel used (now simply retrieve all `UserUttered` and `BotUttered` events from the tracker when you write out a full log)
- the `Dispatcher` gets a `bot_message`, which holds `text` and `data` (`data` can be anything: a button, a generic element, an image)
- when the `run()` method of an action is called, all bot utterances contained in the `Action`'s `Dispatcher` are written to the tracker

- [ ] updated the changelog

### DIFF
--- a/rasa_core/dispatcher.py
+++ b/rasa_core/dispatcher.py
@@ -103,8 +103,6 @@ class Dispatcher(object):
             t["buttons"] = buttons
         else:
             t["buttons"].extend(buttons)
-        self.latest_bot_messages.append(BotMessage(text=t["text"],
-                                                   data={"buttons": t["buttons"]}))
         self.utter_response(t)
 
     def utter_template(self, template, filled_slots=None, **kwargs):
@@ -112,8 +110,6 @@ class Dispatcher(object):
         """"Send a message to the client based on a template."""
 
         message = self.retrieve_template(template, filled_slots, **kwargs)
-        self.latest_bot_messages.append(BotMessage(text=message["text"],
-                                                   data=None))
         self.utter_response(message)
 
     @staticmethod
--- a/rasa_core/dispatcher.py
+++ b/rasa_core/dispatcher.py
@@ -6,6 +6,8 @@ from __future__ import unicode_literals
 import copy
 
 from typing import Text, List, Dict, Any
+from collections import namedtuple
+
 import copy
 
 from rasa_core.domain import Domain
@@ -24,15 +26,7 @@ class Element(dict):
         super(Element, self).__init__(*args, **kwargs)
 
 
-class BotMessage(dict):
-    __acceptable_keys = ['text', 'data']
-
-    def __init__(self, *args, **kwargs):
-        kwargs = {key: value
-                  for key, value in kwargs.items()
-                  if key in self.__acceptable_keys}
-
-        super(BotMessage, self).__init__(*args, **kwargs)
+BotMessage = namedtuple("BotMessage", "text data")
 
 
 class Button(dict):
@@ -50,7 +44,7 @@ class Dispatcher(object):
         self.output_channel = output_channel
         self.domain = domain
         self.send_messages = []
-        self.bot_message = None
+        self.latest_bot_messages = []
 
     def utter_response(self, message):
         # type: (Dict[Text, Any]) -> None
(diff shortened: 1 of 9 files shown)


rasa_core/dispatcher.py
```diff
@@ -39,6 +50,7 @@ def __init__(self, sender_id, output_channel, domain):
         self.output_channel = output_channel
         self.domain = domain
         self.send_messages = []
+        self.bot_message = None
```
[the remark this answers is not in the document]
tmbo: We also need to store multiple messages here, the action might utter multiple messages during the run method.

rasa_core/events/__init__.py
```diff
@@ -148,6 +148,42 @@ def apply_to(self, tracker):
...
 
+# noinspection PyProtectedMember
+class BotUttered(Event):
+    """The bot has said something to the user.
+
+    This class is not used in the story training as it is contained in the
+
+    ```ActionExecuted class```. An entry is made in the ``Tracker``."""
+
+    def __init__(self, text=None, data=None):
```
tmbo: class needs a `type_name` and should implement `_from_parameters` (when the actions are executed remotly we want to give users the possibility to log the bot utterances to rasa core), e.g. something like
`{type: "bot", message: "my message", data: {}}`

rasa_core/processor.py
```diff
@@ -287,10 +287,24 @@ def _run_action(self, action, tracker, dispatcher):
...
         self._log_action_on_tracker(tracker, action.name(), events)
+        self._log_bot_utterances_on_tracker(tracker, dispatcher)
         self._schedule_reminders(events, dispatcher)
 
         return self._should_predict_another_action(action.name(), events)
 
+    def _log_bot_utterances_on_tracker(self, tracker, dispatcher):
+        # type: (DialogueStateTracker, Dispatcher) -> None
+
+        if dispatcher.bot_message is not None:
```
tmbo: also needs to reset the `dispatcher.bot_message` otherwise we will log the message again for the next action.

rasa_core/processor.py
```diff
@@ -287,10 +287,24 @@ def _run_action(self, action, tracker, dispatcher):
             logger.error(e, exc_info=True)
             events = []
         self._log_action_on_tracker(tracker, action.name(), events)
+        self._log_bot_utterances_on_tracker(tracker, dispatcher)
```
tmbo: let's do this before logging the actions (makes a bit more sense, e.g. if the action created the restart event, that one would now be logged before the bot utterances - it makes a bit more sense the other way around).

rasa_core/processor.py
```diff
@@ -286,11 +286,33 @@ def _run_action(self, action, tracker, dispatcher):
...
         self._log_action_on_tracker(tracker, action.name(), events)
         self._schedule_reminders(events, dispatcher)
 
         return self._should_predict_another_action(action.name(), events)
 
+    def _log_bot_utterances_on_tracker(self, tracker, dispatcher):
+        # type: (DialogueStateTracker, Dispatcher) -> None
+
+        if dispatcher.latest_bot_messages:
+            text = []
```
tmbo: let's just log the messages separately (so for each ebtry ub `dispatcher.katest_bot_messages` one event `BotUttered`)  - or is there a reason not to do that?

rasa_core/training_utils/dsl.py
```diff
@@ -128,6 +128,8 @@ def as_story_string(self, flat=False):
         for s in self.events:
             if isinstance(s, UserUttered):
                 result += "* {}\n".format(s.as_story_string())
+            elif isinstance(s, BotUttered):
```
tmbo: Ok either
- change this so that the ignore happens in the `elif isinstance(s, Event)` case
```
elif isinstance(s, Event):
   converted = s.as_story_string()
   if converted:
     result += "    - {}\n".format(converted)
```
- or let `BotUttered.as_story_string()` throw an exception again

Otherwise, there is a `as_story_string` that never gets called.

tmbo:
Nice 👍  Ok, so the code is looking really good, now we need a couple of tests to make sure this doesn't break in the future:
- there are already event tests - BotUttered needs to be added there
- there should be a test that just tests `_log_bot_utterances_on_tracker` (passing in made up arguments and checking that the messages stored on the dispatcher have been reset) 
- another test that contains a storystep with a `BotUttered` event and which tests the call to `as_story_string`

tmbo:
And don't forget to sign the CLA :wink:

ricwo:
cool, thanks :)

the main changes are:
- added test for `BotUttered` events
- testing `log_bot_utterances_on_tracker` (with a `default_processor` for tests in general)
- `as_story_string` is tested as before (where `None` is excluded), as well as with the `BotUttered` event returning `None`

PR #107, opened 2017-11-20
Fix json cannot be decoded when OPTIONS is called before POST

felixcheruiyot:
Calling a post with continue endpoint from a browser client I get the following error
exceptions.ValueError: No JSON object could be decoded

This fix is intended to ensure the endpoint if only fully executed when a POST method is called.

**Proposed changes**:
- ...

- [ ] updated the changelog


tmbo:
Thanks for the contribution. I fixed it a bit differently as this code is also used in Rasa NLU and I needed a general solution that avoids adding an `if` to every route.

PR #159, opened 2018-01-05
timestamps

amn41:
**Proposed changes**:
- adds timestamp for `UserUttered` and `BotUttered` events
- adds `latest_message_time` attribute to tracker

- [ ] updated the changelog

### DIFF
--- a/rasa_core/events/__init__.py
+++ b/rasa_core/events/__init__.py
@@ -33,8 +33,8 @@ class Event(object):
 
     type_name = "event"
 
-    def __init__(self):
-        self.time = time.time()
+    def __init__(self, timestamp=None):
+        self.timestamp = timestamp if timestamp else time.time()
 
     def __ne__(self, other):
         # Not strictly necessary, but to avoid having both x==y and x!=y
@@ -57,10 +57,13 @@ class Event(object):
     @classmethod
     def _from_story_string(cls, event_name, parameters, domain):
         """Called to convert a parsed story line into an event."""
-        return cls()
+        return cls(parameters.get("timestamp"))
 
     def as_dict(self):
-        return {"event": self.type_name}
+        return {
+            "event": self.type_name,
+            "timestamp": self.timestamp,
+        }
 
     @classmethod
     def _from_parameters(cls, event_name, parameters, domain):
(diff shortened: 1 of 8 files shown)


rasa_core/trackers.py
```diff
@@ -79,10 +79,16 @@ def current_state(self, should_include_events=False):
         else:
             events = None
 
+        latest_message_time = None
+        timestamps = [e.time for e in self.events if hasattr(e,'time')]
+        if len(timestamps) > 0:
+            latest_message_time = max(timestamps)
```
tmbo: is there any case where the events are not sorted by timestamps?

rasa_core/events/__init__.py
```diff
@@ -98,6 +99,7 @@ def __init__(self, text, intent=None, entities=None, parse_data=None):
         self.text = text
         self.intent = intent if intent else {}
         self.entities = entities if entities else []
+        self.time = time.time()
```
tmbo: I think we talked about this already, but whats the reason to not do this at the event level, so for every event?

amn41: yeah that would be better.

maybe: define an `__init__` method for the `Event` class and then in each of the subclasses, at the end of `__init__` call `super().__init__()` ?

rasa_core/trackers.py
amn41: ah you mean take the last of the events rather than the max? no I suppose there's no reason they wouldn't be sorted

rasa_core/events/__init__.py
tmbo: yes

rasa_core/trackers.py
tmbo: cool - if we add the time to all events the last one should have the timestamp :+1:

amn41:
@tmbo the failing build is due to an error in the `docs` stage

tests/test_server.py
```diff
@@ -9,6 +9,7 @@
 from builtins import str
 
 import pytest
+from freezegun import freeze_time
```
amn41: of course there's a package for this

tmbo: its python there is a package for everything 😍

PR #173, opened 2018-01-15
Don't fail on unkown actions

tmbo:
**Proposed changes**:
- e.g. if you remove an action from the domain, logs from the past might still contain that action. so we shouldn't fail here but rather just warn
- for the future we should think wether it makes sense to remove any samples that contain actions that are not part of the domain anymore.

- [x] updated the changelog

### DIFF
--- a/rasa_core/domain.py
+++ b/rasa_core/domain.py
@@ -288,7 +288,7 @@ class Domain(with_metaclass(abc.ABCMeta, object)):
                         "Please make sure all actions are listed in the "
                         "domains action list. If you recently removed an "
                         "action, don't worry about this warning. It "
-                        "should stop to appear after a while. "
+                        "should stop appearing after a while. "
                         "".format(latest_action))
         else:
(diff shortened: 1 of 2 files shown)


rasa_core/domain.py
```diff
@@ -283,10 +283,13 @@ def get_prev_action_features(self, tracker):
...
                 return {prev_action_name: 1}
             else:
-                raise Exception(
+                logger.warn(
                         "Failed to use action '{}' in history. "
                         "Please make sure all actions are listed in the "
-                        "domains action list.".format(latest_action))
+                        "domains action list. If you recently removed an "
+                        "action, don't worry about this warning. It "
+                        "should stop to appear after a while. "
```
amn41: "should stop appearing"

tmbo: :+1:

tmbo:
@amn41 message updated.

PR #178, opened 2018-01-17
Update structures.py

abinj:
Removed the non required type 'Deque' from structures.py

**Proposed changes**:
- ...

- [ ] updated the changelog


tmbo:
Thanks for your PR. Unfortunately, we can not simply remove that type as it is used for the type annotations (take a look at L257 in that file).

abinj:
Hello Tom,

I understood, it was my mistake.



Thanks

PR #185, opened 2018-01-22
Bug fix: utter_custom_message never called (to display carousels in Messenger)

znat:
**Proposed changes**:
`utter_custom_message ` is never called on the `Dispatcher` to display carousel objects in Messenger. I just added a condition that looks for an `elements` key in the message and executes `utter_custom_message ` if it finds it

- [ ] updated the changelog


tmbo:
Very good point :rocket:

Issue #193, opened 2018-01-25
UserUtteranceReverted event incorrectly replays actions


jctrouble:
**Rasa Core version**:
master (5a0f0ea445da8f6cbfe53402aa70b1900fd23f66)

**Python version**: 
3.6.3

**Operating system** (windows, osx, ...):
OSX 10.13.2

## Example
Sample project illustrating the error:
https://github.com/twineai/rasa_core-1/tree/rewind_error/examples/rewindbot

![screen shot 2018-01-25 at 12 51 43 pm](https://user-images.githubusercontent.com/1136226/35411703-8e627c7e-01ce-11e8-875a-3d27f26ce240.png)

## Issue
In the example above, the `do_inconsequential_thing` action issues a `UserUtteranceReverted` event. After the event is applied, the bot predicts the next actions to be the actions that were previously ran, whereas I'd expect the bot's state to be simply as if the utterance never occurred at all.

## Diagnosis
I believe it's the result of the following:
*Prior to utterance*
```
Tracker Event History =
   - ActionExecuted(Listen)
   - UserUttered(do_destructive_thing)
   - ActionExecuted(do_destructive_thing_one)
   - ActionExecuted(do_destructive_thing_two)
```

*At this moment, the bot is in the process of listening*
*Now the utterance comes in, and the next action runs (the one which yields the `UserUtteranceReverted`)*
```
Tracker Event History =
   - ActionExecuted(Listen)
   - UserUttered(do_destructive_thing)
   - ActionExecuted(do_destructive_thing_one)
   - ActionExecuted(do_destructive_thing_two)
   - ActionExecuted(listen)
   - UserUttered(do_inconsequential_thing)
   - ActionExecuted(do_inconsequential_thing)
```

*The `UserUtteranceReverted` then rewinds the second `UserUttered`, yielding a state like this:*
```
Tracker Event History =
   - ActionExecuted(Listen)
   - UserUttered(do_destructive_thing)
   - ActionExecuted(do_destructive_thing_one)
   - ActionExecuted(do_destructive_thing_two)
   - ActionExecuted(listen)
```

*Then the policy looks at the state and says "what do I do in this situation", the model sees the last **completed** action (this is important) being the listen, and tries to recommend an action. This queries the last message for the intent (`do_destructive_thing` in this case), and then runs the associated actions.*

[...]

PR #194, opened 2018-01-25
Fix UserUtteranceReverted event to include the prior listen

jctrouble:
**Proposed changes**:
- When rewinding events due to a `UserUtteranceReverted` event, also rewind the previous action, which is by definition a listen

**Note:**
This seemed like the easy fix, but there may be a more elegant solution. Happy to look at alternate approaches.

- [n/a] updated the documentation
- [x] updated the changelog


jctrouble:
this is to fix #193

PR #195, opened 2018-01-26
Adds conflict detector for training data

jctrouble:
**Proposed changes**:
Adds an optional argument to `Agent.train()` which will emit warnings if the generated training data contains conflicting labels for the same feature vector.

This is probably not the most efficient way to go about it, so I left it off by default, but it can come in handy during those "why is my accuracy so low?" moments.

- [n/a] added some tests for the functionality
- [x] updated the changelog


tmbo:
@jctrouble Sorry for not getting back to you on this PR. 

I think the idea is very valuable and I think it will make issue detection quite a bit easier. The issue I see with the current implementation is that there is no help in finding the conflicting samples.

PR #199, opened 2018-01-29
Updating common patterns documentation

mlennox:
**Proposed changes**:
- added section highlighting the use of `Slot` in stories

- [ ] updated the changelog

### DIFF
--- a/docs/patterns.rst
+++ b/docs/patterns.rst
@@ -50,11 +50,15 @@ this will be saved as a slot,
 
 When Rasa Core predicts the next action to take, the only information it has about the ``TextSlot`` s is 
 **whether or not they are defined**. So you have enough information to know that you don't have to ask for this
-information again, but the *value* has no impact on which actions Rasa Core predicts.
+information again, but the *value* of a ``TextSlot`` has no impact on which actions Rasa Core predicts. This is
+explained in :ref:`more detail below <slot_features>`.
+
 The full set of slot types and their behaviour is described here: :ref:`slot_types`.
 
 Using Slot Values to Influence Which Actions are Predicted
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
+
+**Custom Slots**
 
 Maybe your restaurant booking system can only handle bookings for up to 6 people, so the 
 request above isn't valid. In this case you want the *value* of the slot to influence the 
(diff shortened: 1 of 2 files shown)


amn41:
thanks! please mark as 'ready for review' when you want us to take a look.

mlennox:
I'll work on the Checkpoints docs in a separate PR I think

docs/patterns.rst
```diff
@@ -114,6 +118,86 @@ a ``categorical`` slot.
...
+        values:
+        - unknown
+        - booked-out
+        - waiting-list
+        - available
+
+When the ``restaurant_availability`` slot is set Rasa Core will be able to determine if the restaurant in question is
+available and choose radically different actions to perform based on the value.
+
+The ``Slot`` *might* be set by Rasa Core itself from entities detected by the NLU module, but usually you would
```
amn41: let's not say *might*, but rather be explicit. If NLU detects this as an entity, and there is a slot with the same name, this value will be set as a slot.

docs/patterns.rst
```diff
@@ -114,6 +118,86 @@ a ``categorical`` slot.
...
+        - unknown
+        - booked-out
+        - waiting-list
+        - available
+
+When the ``restaurant_availability`` slot is set Rasa Core will be able to determine if the restaurant in question is
+available and choose radically different actions to perform based on the value.
+
+The ``Slot`` *might* be set by Rasa Core itself from entities detected by the NLU module, but usually you would 
+return the value of the ``Slot`` from your ``Action`` and then use the next ``Turn`` of the conversation to
```
amn41: I would mention here that the primary use case for returning SlotSet events in an action is to collect information about the outside world (e.g. by making an API call) and feeding that information back into the dialogue.

docs/patterns.rst
```diff
@@ -114,6 +118,86 @@ a ``categorical`` slot.
...
+check what feature is set.
+
+.. code-block:: python
+
+    def run(self, dispatcher, tracker, domain):
+        # some logic here to decide if the restaurant is "available", "booked-out" or whatever
+        return [SlotSet("restaurant_availability", "available")]
+
+In this first story we will try and make a booking for 5 people in a restaurant on the night of 21st August 2018. 
+In this case the restaurant is booked out so we want to apologise to the customer and suggests similar restaurants.
```
amn41: Make a comment here that writing these stories out by hand, and they are shown for illustration. Using interactive learning to generate these stories is *much* less error-prone

docs/patterns.rst
amn41: here either refer to the last section  "Storing API responses in the tracker" or just include that seciton here

docs/patterns.rst
mlennox: agreed

docs/patterns.rst
mlennox: interesting, that makes sense, but I hadn't realised it made such a difference

docs/patterns.rst
```diff
@@ -114,15 +118,119 @@ a ``categorical`` slot.
...
+
+    restaurant_availability:
+        type: categorical
+        values:
+        - unknown
+        - booked-out
+        - waiting-list
+        - available
+
+A ``Slot`` will be set by Rasa Core if its name and the name of the entity detected by the NLU module match. However, when we wish to alter the flow of a conversation we would explicitly set the value of the slot from a custom ``Action`` and
```
amn41: this isn't quite right. whether the slot is set automatically or in a custom action, it has the same ability to impact the conversation flow. So long as you have relevant training data, rasa core will pick up on these patterns

docs/patterns.rst
```diff
@@ -114,15 +118,119 @@ a ``categorical`` slot.
...
+        def run(self, dispatcher, tracker, domain):
+            restaurant_name=tracker.get_slot("restaurant_name")
+            location=tracker.get_slot("location")
+            num_people=tracker.get_slot("people")
+            date=tracker.get_slot("date")
+            # this will fetch the availability of the restaurant from your DB or an API
+            availability=restaurantService.check_availability(restaurant_name, location, num_people, date)
+            return [SlotSet("restaurant_availability", availability)]
+
+The snippet of code above from a hypothetical ``Action`` shows that the value of the slot ``restaurant_availability`` is determined by querying a database or API. The restaurant availability is not something that is known when we train the dialogue model, the ``Slot`` value is the only way we can alter the course of the conversation based on information from the outside world.
```
amn41: awesome! this is very clearly explained - I think the example helps a lot

docs/patterns.rst
```diff
@@ -114,15 +118,119 @@ a ``categorical`` slot.
...
+
+The data fetched from an API call can also be stored for later use without altering the outcome of a conversation as detailed in :ref:`unfeaturized_slots`.
+
+**Slot Features Example**
+
+.. note:: 
+    These example stories have been constructed manually for illustrative purposes. While this is a valid approach to training your model the preferred approach is to use :ref:`supervised learning <tutorial_supervised>` which generates stories that are *much* less error-prone.
```
amn41: reference should be to *interactive learning*

docs/patterns.rst
```diff
@@ -114,15 +118,119 @@ a ``categorical`` slot.
...
+
+In this first story we will try and make a booking for 5 people in a restaurant on the night of 21st August 2018. 
+In this case the restaurant is booked out so we want to apologize to the customer and suggests similar restaurants. It is assumed that the Rasa Core model has been trained to recognise a message like *"Book Murphys Bistro on August 21 for 5 people"*
```
amn41: and suggests -> and suggest

docs/patterns.rst
```diff
@@ -114,15 +118,119 @@ a ``categorical`` slot.
...
+
+.. code-block:: md
+
+    # restaurant unavailable
+    * _make_booking[people=5, date="2018-08-21T19:30:00+00:00", restaurant_id=145]
```
amn41: we have an updated format format for stories, https://core.rasa.ai/stories.html
so e.g. 
`* make_booking{"people":"5", "date":"2018-08-21T19:30:00+00:00", "restaurant_id":"145"}`

mlennox:
thank you @amn41 I'm implementing those changes now.

tmbo:
Very good :+1: Let me know when you are ready.

mlennox:
@tmbo I'm happy if you are - anything else you need me to do?

PR #208, opened 2018-02-03
Datetime serialisation fix

amn41:
**Proposed changes**:
- this fixes a bug trying to JSON serialise stories containing `ReminderScheduled` events. 

Would throw an exception like:
`TypeError: datetime.datetime(2018, 2, 3, 16, 32, 10, 827475) is not JSON serializable`

- [x] updated the changelog


rasa_core/events/__init__.py
```diff
@@ -475,7 +475,7 @@ def __str__(self):
     def as_story_string(self):
         props = json.dumps({
             "action": self.action_name,
-            "date_time": self.trigger_date_time,
+            "date_time": self.trigger_date_time.isoformat(),
```
tmbo: Need to parse this in the `_from_story_string` function again.

tmbo:
Can you also please add a test for that function?

tests/test_events.py
```diff
@@ -68,3 +68,34 @@ def test_event_has_proper_implementation(one_event, another_event):
...
+
+    ActionExecuted("my_action"),
+
+    BotUttered("my_text", "my_data"),
+
+    ReminderScheduled("my_action", "now")
+])
+def test_serialisation(one_event):
+    story_string = one_event.as_story_string()
+    new_event = Event.from_story_string(story_string, one_event.as_dict(), None)
```
amn41: @tmbo should this work for all events? (does for reminderscheduled) - do we need a real `domain` for any of these?

tmbo: Yes, I think at least `ActionExecuted` needs it. best is to add `default_domain` as a parameter - will be filled automatically and use that one. The default_domain also has a slot `name` that you can set to a value. Best is to also use an action that is in there, so e.g. `utter_greet`.

rasa_core/events/__init__.py
tmbo: ✅

tests/test_events.py
amn41: alright will update that - didn't work for `TopicSet` but that might also be related to domain

PR #218, opened 2018-02-13
moved polciy to own model to allow it getting persisted properly

tmbo:
**Proposed changes**:
- moved policy to own model to allow it getting persisted properly (otherwise it will get persisted with `__main__` as a module path, which can not be loaded consistently)

- [ ] updated the changelog

### DIFF
--- a/docs/tutorial_supervised.rst
+++ b/docs/tutorial_supervised.rst
@@ -151,9 +151,9 @@ For this bot, we came up with our own policy.
 This policy extends the Keras policy modifying the ML architecture of the
 underlying neural network.
 Check out the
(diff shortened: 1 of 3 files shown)


docs/tutorial_supervised.rst
```diff
@@ -151,9 +151,9 @@ For this bot, we came up with our own policy.
 This policy extends the Keras policy modifying the ML architecture of the
 underlying neural network.
 Check out the
-``RestaurantPolicy`` class in ``bot.py`` for the gory details:
```
Ghostvv: I thought `gory` was used on purpose

PR #229, opened 2018-02-16
Project in http nlu interpreter

znat:
**Proposed changes**:
The main change (http params) is adding project name as a parameter for http interpreter

The other change (`basestring`) was initially for another PR but branches got merged... It fixes a compatibility issue with 2.7: when running core with a custom script and passing the domain as a parameter:
```
python myscript.py -d domain.yml
```
And if the shell passes args in unicode, `if isinstance(domain, str):` will crash because unicode is not an instance of `str`

- [ ] updated the changelog


rasa_core/agent.py
```diff
@@ -233,7 +233,7 @@ def _create_featurizer(cls, featurizer):
 
     @classmethod
     def _create_domain(cls, domain):
-        if isinstance(domain, str):
+        if isinstance(domain, basestring):
```
tmbo: let's do 
```
from six import string_types

if isinstance(domain, string_types):
```
`basestring` is not defined in py3

PR #238, opened 2018-02-23
implementation of support for loading stories from multiple files

tmbo:
**Proposed changes**:
- instead of specifying a single file for stories, this allows to load stories from a directory of story files
- in the future we can think about specifying multiple files directly (instead of a folder) - but this was easier to do for now

- [x] updated the changelog

### DIFF
--- a/rasa_core/training/dsl.py
+++ b/rasa_core/training/dsl.py
@@ -8,10 +8,10 @@ import io
 import json
 import logging
 import os
-import rasa_nlu
 import re
 import warnings
 
+from rasa_nlu import utils as nlu_utils
 from typing import Optional, List, Text, Any, Dict
 
 from rasa_core import utils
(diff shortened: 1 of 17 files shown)


rasa_core/training/dsl.py
```diff
@@ -8,6 +8,7 @@
 import json
 import logging
 import os
+import rasa_nlu
```
amn41: we need to specifically import this submodule, let's do:
`from rasa_nlu import utils as nlu_utils`

Ghostvv:
trainer also reads hidden file on my mac .DS_Store

hotzenklotz:
Very nice :-) I am looking forward to this.

tmbo:
I've address the hidden file issue & your comment @amn41 

mind taking another look?

tmbo:
https://github.com/RasaHQ/rasa_nlu/pull/904 needs to be merged into nlu before this can be merged.

PR #261, opened 2018-03-12
Form fill

amn41:
**Proposed changes**:
- addresses #242 
- adds a `FormAction` class which implements a simple logical loop to collect some slots before submitting
- rather than having separate actions for each question, users define templates for `utter_ask_{slot_name}` for each slot, and rasa core should repeatedly call the `FormAction`, e.g.
```
I'd like to book a restaurant
 - sure for how many people?
3
  - what time?
 8pm
 - ok it's booked
```
would look like:
```
* request_restaurant
     - action_restaurant_form
     - slot{"requested_slot": "people"}
* inform{"number": 3}
     - action_restaurant_form
     - slot{"people": 3}
     - slot{"requested_slot": "time"}
* inform{"time": "8pm"}
   - action_restaurant_form
...
```

- [ ] updated the changelog


hotzenklotz:
I am really looking forward to simplifying from interactions. So this PR is awesome. I have a question about the story format, though. Don't you have to include the `SlotSet` events of the custom action after every call to an action? Otherwise it can't be learned by the LSTM? E.g.

```
* request_restaurant
   - action_restaurant_form
   - slot{"restaurant": "not_null"}
* inform{"number": 3}
   - action_restaurant_form
   - slot{"number": "not_null"}
...
```

Also, do you intent to include a mechanism to include information given ahead of time. E.g. "I want to eat at an Italian restaurant for 3 people." Only info about the time would be missing from the form flow now.

amn41:
yes you are totally right about the slots! goodness typing those manually is really a trap. we actually just caught this applying it elsewhere. 

the case of "I want to eat an an italian restaurant with 3 people" should  already work (provided there's a story) because the corresponding slots will be set. But this relies on your NER being able to pick up "cuisine" and "number_of_people" slots. This is easy in the restaurant case, but when you prompt a user for an answer, and they just type a number, you don't necessarily know what slot that should be set as.

hotzenklotz:
> yes you are totally right about the slots! goodness typing those manually is really a trap. we actually just caught this applying it elsewhere.

Our customers' feedback was an equally shocked reaction when we told them to define all theses slots manually ;-)

amn41:
interactive learning is a *much* less error-prone way to generate stories

amn41:
cool @tmbo @akelad I updated docs as well please have a look

docs/patterns.rst
```diff
@@ -15,8 +15,71 @@ Slots are for storing information that's relevant over multiple turns. For examp
...
+
+Some important things to consider:
+
+- Your domain needs to have a slot called ``requested_slot``. This can be unfeaturized, but if you want
+  to support contextual questions like *"why do you need to know that information?"* you should make this
+  a categorical slot. 
+- You need to define utterances for asking for each slot in your domain, e.g. ``utter_ask_{slot_name}``.
+- We strongly recommend that you create these stories using interactive learning, because if you
+  type these by hand you will probably forget to add all of the slots.
+- Any slots that are already set won't be asked for. E.g. if someone says "I'd like a Chinese restaurant for 8 people" the ``submit`` function should get called right away.
```
akelad: would it make sense to briefly comment on the use of the `validate` method as well?

rasa_core/actions/forms.py
```diff
@@ -0,0 +1,111 @@
...
+                dispatcher.utter_template("utter_ask_{}".format(field.slot_name))
+
+                events.append(SlotSet("requested_slot", field.slot_name))
+                return events
+
+        return self.submit(dispatcher, tracker, domain)
+
+
+    def submit(self, dispatcher, tracker, domain):
+        dispatcher.utter_message("done!")
```
akelad: Do we want the default to be the bot saying "done!" ? or maybe just a message in the logger saying "Finished filling slots"

docs/patterns.rst
```diff
@@ -15,8 +15,71 @@ Slots are for storing information that's relevant over multiple turns. For examp
...
+
+A form action has a set of required fields, which you define for the class:
+
+.. code-block:: python
+
+   class RestaurantSearch(FormAction):
+   
+       REQUIRED_FIELDS = [
+	  EntityFormField("number", "num_people"),
+          EntityFormField("cuisine", "cuisine")
```
tmbo: not aligned

rasa_core/actions/forms.py
```diff
@@ -0,0 +1,111 @@
...
+
+class EntityFormField(FormField):
+
+    def __init__(self, entity_name, slot_name):
+        self.entity_name = entity_name
+        self.slot_name = slot_name
+
+    def extract(self, tracker):
+        value = None
+        for e in tracker.latest_message.entities:
```
tmbo: `value = next(tracker.get_latest_entity_values(self.entity_name), None)`

rasa_core/actions/forms.py
```diff
@@ -0,0 +1,111 @@
...
+class BooleanFormField(FormField):
+
+    def __init__(self, slot_name, affirm_intent, deny_intent):
+        self.slot_name = slot_name
+        self.affirm_intent = affirm_intent
+        self.deny_intent = deny_intent
+
+    def extract(self, tracker):
+        value = None
+        intent = tracker.latest_message.intent["name"]
```
tmbo: should be `.get("name")` might be `None`

rasa_core/actions/forms.py
```diff
@@ -0,0 +1,111 @@
...
+        return not any([
+            self.should_request_slot(tracker, field.slot_name, events)
+            for field in self.REQUIRED_FIELDS])
+
+
+    def run(self, dispatcher, tracker, domain):
+
+        events = self.get_requested_slot(tracker)
+
+        if self.ready_to_submit(tracker, events):
```
tmbo: do we need this?

I mean there is a submit at the end of the function anyways

rasa_core/actions/forms.py
```diff
@@ -0,0 +1,111 @@
...
+        if requested_slot is None:
+            return []
+        else:
+            try:
+                required = self.REQUIRED_FIELDS[:]
+                if self.RANDOMIZE:
+                    random.shuffle(required)
+                fields = [f for f in required if f.slot_name == requested_slot]
+                return fields[0].extract(tracker)
+            except:
```
tmbo: whats the point in catching exceptions and raising them again?

rasa_core/actions/forms.py
```diff
@@ -0,0 +1,111 @@
...
+    def get_requested_slot(self, tracker):
+        requested_slot = tracker.get_slot("requested_slot")
+
+        if requested_slot is None:
+            return []
+        else:
+            try:
+                required = self.REQUIRED_FIELDS[:]
+                if self.RANDOMIZE:
+                    random.shuffle(required)
```
tmbo: what is the point in shuffeling, isn't there only one element in `requested_slot` anyways?

rasa_core/actions/forms.py
```diff
@@ -0,0 +1,111 @@
...
+                fields = [f for f in required if f.slot_name == requested_slot]
+                return fields[0].extract(tracker)
```
tmbo: please check that there is at least one field

rasa_core/actions/forms.py
amn41: it shuffles the 'required' slots not the requested one. So if you want, the questions don't always come in the same order

rasa_core/actions/forms.py
amn41: ah right, idea behind this was that users could override this, in case some of the 'required_slots' are actually optional (e.g. you don't need all the slots to search for a restaurant)

tmbo:
the things I am concerned about:
- using the same action for the whole form filling makes the stories a lot less readable (as the whole state of the form is hidden within the action)
- because it is always the same action name, setting the right slots is even harder by hand

amn41:
I agree about the actions making things less readable (but the slots may help there). Slots are always easy to forget, but I'm not sure if they are easier or harder here? 

Maybe we should create a script to sanity-check story files by running through the actions and checking if any slots are missing

kristianmitk:
We just tested it and its doing the job well!  🚀 

But afaiu using this with `factory_remote` won't work, or?

Any way this would be executed even though the http mode is activated and the server returns whatever `utter_ask_{slot_name}`  as the `next_action`?

amn41:
glad to hear it @kristianmitk ! Hmm no this wouldn't work over HTTP, you would have to implement something similar to the `FormAction` in your own code

AbdurRub:
when should we expect merging of this feature? @amn41

akelad:
@amn41 just fyi the randomisation of the order of asking for slots doesn't work the way it's currently implemented. I know how to fix it though - i'll push that soon

amn41:
API based slot validation:
- allow users to write custom logic for whether an entity becomes a slot, and *which* entities become slots
- allow for free-text input

amn41:
@akelad just a reminder for your randomization fix

akelad:
yes thanks for reminding me, i keep forgetting

tmbo:
@amn41 is this ready for another review?

amn41:
@tmbo yes ! 

I got rid of the `ready_to_submit` thing bc if users want to customise that they can just override the `run()` method of the action.

rasa_core/actions/forms.py
```diff
@@ -0,0 +1,108 @@
...
+        required = self.REQUIRED_FIELDS[:]
+
+        if requested_slot is None:
+            return []
+        else:
+            fields = [f for f in required if f.slot_name == requested_slot]
+            if len(fields) > 0:
+                return fields[0].extract(tracker)
+            else:
+                logger.debug("unable to extract value "
```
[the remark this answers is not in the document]
amn41: debatable!
I can see your point, but if anything it should be a warning shown to the user of the chatbot, not the developer. The form is working as expected, no?

rasa_core/actions/forms.py
```diff
@@ -0,0 +1,108 @@
...
+            if self.should_request_slot(tracker, field.slot_name, events):
+                dispatcher.utter_template("utter_ask_{}".format(field.slot_name))
+
+                events.append(SlotSet("requested_slot", field.slot_name))
+                return events
+
+        return self.submit(dispatcher, tracker, domain, events)
+
+    def submit(self, dispatcher, tracker, domain, events):
+        dispatcher.utter_message("done!")
```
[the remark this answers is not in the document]
amn41: well the idea is that this is where you actually _do something_, so for a useful `FormAction` the developer has to override this anyway

rasa_core/actions/forms.py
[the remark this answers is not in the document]
tmbo: yes that makes sense, it is not very helpful for a developer, but might hint at wrong training data

rasa_core/actions/forms.py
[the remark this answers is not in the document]
tmbo: Than we need to make that explicit I think. Throwing a `NotImplementedError` should be fine adding a short description to the function.

rasa_core/actions/forms.py
```diff
@@ -0,0 +1,108 @@
...
+
+class FormAction(Action):
+
+    REQUIRED_FIELDS = []
+    RANDOMIZE = True
+
+    def should_request_slot(self, tracker, slot_name, events):
+        existing_val = tracker.get_slot(slot_name)
+        pending = [e.key for e in events if e.key == slot_name]
+        return existing_val is None and slot_name not in pending
```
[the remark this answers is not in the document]
amn41: yes good catch! I think the way I've refactored it, it should work now. 

the only thing I really don't like, is that the user's `submit` method takes in events as well, and has to return them, appending any custom events. for example, this car quote calculation:

```
    def submit(self, dispatcher, tracker, domain, events):
        dispatcher.utter_message("getting your quote")
        quote = "$100"
        dispatcher.utter_message("it's $100")
      	_events = events or []
        return _events +	[SlotSet("quote", quote)]
```

I don't like that solution very much, maybe we can abstract that part away

praveenrengarajan:
have you done documentation for this form fill features?

rasa_core/actions/forms.py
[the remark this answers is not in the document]
tmbo: can't we do the concatenation of the events in the `run` method? submit returns events and they get concatenated in the run method. do you need the `events` in the submit method?

amn41:
@hotzenklotz if you could give this another spin that would be 💯

amn41:
FYI  @tmbo added a `copy()` method to tracker for this purpose

rasa_core/actions/forms.py
```diff
@@ -0,0 +1,138 @@
...
+
+        for field in self.required_fields():
+            if self.should_request_slot(tracker, field.slot_name, events):
+                dispatcher.utter_template(
+                    "utter_ask_{}".format(field.slot_name))
+
+                events.append(SlotSet("requested_slot", field.slot_name))
+                return events
+
+        submit_tracker = tracker.copy()
```
tmbo: if we move this up before the `for` we don't need to do the `pending` stuff in `should_request_slot` anymore

tmbo:
🎉

Issue #119, opened 2017-11-27
Simple Rasa Core + NLU Rest API delivering responses


jenniferzhu:
My goal is to wrap the rasa core + nlu bot into an API so that I can call it in some JavaScript-based UI, such as a simple angularjs-chat ( http://angularjs.chat/tutorials/angularjs-basic-chat-module/).

 However, I can only use CustomInput to send a message to "localhost:5005" by POST but I cannot get the response from my bot. Is there any way that I could build/modify on top of the rasa_core? I know that it is already using Flask, so I don’t want to build another Flask app on top of that.

In summary, I guess it would be great if we can set up an API such as "localhost:5005" to receive the questions from users and send responses back to the same channel, like what we have using the consoleInput.

tmbo:
You can create a custom input channel like this:

```python
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function
from __future__ import unicode_literals

import logging

from flask import Blueprint, request, jsonify

from rasa_core.channels.channel import UserMessage
from rasa_core.channels.direct import CollectingOutputChannel
from rasa_core.channels.rest import HttpInputComponent

logger = logging.getLogger(__name__)


class SimpleWebBot(HttpInputComponent):
    """A simple web bot that listens on a url and responds."""

    def blueprint(self, on_new_message):
        custom_webhook = Blueprint('custom_webhook', __name__)

        @custom_webhook.route("/", methods=['GET'])
        def health():
            return jsonify({"status": "ok"})

        @custom_webhook.route("/webhook", methods=['POST'])
        def receive():
            payload = request.json
            sender_id = payload.get("sender", None)
            text = payload.get("message", None)
            out = CollectingOutputChannel()
            r = on_new_message(UserMessage(text, out, sender_id))
            responses = [m for _, m in out.messages]
            return jsonify(responses)

        return custom_webhook
```

A request to `/webhook` will give a json response like `["How are you?"]` with the messages of the bot.

schafferer:
Hey, @tmbo, i tried this approach, and it just returns an empty list as response. The bot works just fine from the command line. I see the response gets stored i the tracker object, but there is no messages in out.messages.

sachin-srijan:
hey @tmbo  same here, there is no proper documentation how to use that custom channel.

pradeepmvn:
There is a validation error when I execute this custom channel

[...]

PR #262, opened 2018-03-12
add direct endpoint to rasa core http server

amn41:
**Proposed changes**:
- addresses https://github.com/RasaHQ/rasa_core/issues/119

- [ ] updated the changelog

### DIFF
--- a/rasa_core/channels/direct.py
+++ b/rasa_core/channels/direct.py
@@ -16,4 +16,7 @@ class CollectingOutputChannel(OutputChannel):
             return None
 
     def send_text_message(self, recipient_id, message):
-        self.messages.append((recipient_id, message))
+        self.messages.append({"recipient_id": recipient_id, "text": message})
+
+    def send_text_with_buttons(self, recipient_id, message, buttons):
+        self.messages.append({"recipient_id": recipient_id, "text": message, "data": buttons})
--- a/rasa_core/server.py
+++ b/rasa_core/server.py
@@ -322,7 +322,7 @@ class RasaCoreServer(object):
         except Exception as e:
             request.setResponseCode(500)
             logger.error("Caught an exception during "
-                         "parse: {}".format(e), exc_info=1)
+                         "respond: {}".format(e), exc_info=1)
             return json.dumps({"error": "{}".format(e)})
 
     @app.route("/load", methods=['POST', 'OPTIONS'])
--- a/rasa_core/server.py
+++ b/rasa_core/server.py
@@ -106,6 +106,23 @@ def bool_arg(request, name, default=True):
     return request.args.get(name, d).lower() == 'true'
 
 
+def request_parameters(request):
+    if request.method.decode('utf-8', 'strict') == 'GET':
+        return {
+            key.decode('utf-8', 'strict'): value[0].decode('utf-8',
+                                                           'strict')
+            for key, value in request.args.items()}
+    else:
+        content = request.content.read()
+        try:
+            return json.loads(content.decode('utf-8', 'strict'))
+        except ValueError as e:
+            logger.error("Failed to decode json during respond request. "
+                         "Error: {}. Request content: "
+                         "'{}'".format(e, content))
+            raise
+
+
 class RasaCoreServer(object):
     """Class representing a Rasa Core HTTP server."""
 
@@ -257,20 +274,7 @@ class RasaCoreServer(object):
     @ensure_loaded_agent
     def parse(self, request, sender_id):
         request.setHeader('Content-Type', 'application/json')
-        if request.method.decode('utf-8', 'strict') == 'GET':
-            request_params = {
-                key.decode('utf-8', 'strict'): value[0].decode('utf-8',
-                                                               'strict')
-                for key, value in request.args.items()}
-        else:
-            content = request.content.read()
-            try:
-                request_params = json.loads(content.decode('utf-8', 'strict'))
-            except ValueError as e:
-                logger.error("Failed to decode json during parse request. "
-                             "Error: {}. Request content: "
-                             "'{}'".format(e, content))
-                raise
+        request_params = request_parameters(request)
 
         if 'query' in request_params:
             message = request_params.pop('query')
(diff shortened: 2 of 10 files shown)


rasa_core/channels/direct.py
```diff
@@ -16,4 +16,7 @@ def latest_output(self):
             return None
 
     def send_text_message(self, recipient_id, message):
-        self.messages.append((recipient_id, message))
+        self.messages.append({"recipient_id": recipient_id, "text": message})
```
amn41: @tmbo before I go update a bunch of tests - does this make sense to change? idea is for collectingoutputchannel to collect full messages with a `data` attribute, not just strings

tmbo: sounds good :+1

tmbo:
Is this ready for review?

amn41:
yes- thanks!

rasa_core/server.py
```diff
@@ -288,6 +290,50 @@ def parse(self, request, sender_id):
...
 
+    @app.route("/conversations/<sender_id>/respond",
+               methods=['GET', 'POST', 'OPTIONS'])
+    @check_cors
+    @requires_auth
+    @ensure_loaded_agent
+    def respond(self, request, sender_id):
+        request.setHeader('Content-Type', 'application/json')
+        if request.method.decode('utf-8', 'strict') == 'GET':
+            request_params = {
```
tmbo: please put into separate function and reuse between this and `/parse`

rasa_core/server.py
```diff
@@ -288,6 +290,50 @@ def parse(self, request, sender_id):
...
+                key.decode('utf-8', 'strict'): value[0].decode('utf-8',
+                                                               'strict')
+                for key, value in request.args.items()}
+        else:
+            content = request.content.read()
```
tmbo: same here

tmbo: maybe one function for the whole block is fine, e.g. `request_params = request_parameters(request)`

rasa_core/server.py
```diff
@@ -288,6 +290,50 @@ def parse(self, request, sender_id):
...
+            message = request_params.pop('query')
+        elif 'q' in request_params:
+            message = request_params.pop('q')
+        else:
+            request.setResponseCode(400)
+            return json.dumps({"error": "Invalid parse parameter specified"})
+
+        try:
+            out = CollectingOutputChannel()
+            processor = self.agent._create_processor()
```
tmbo: let's directly use `self.agent.handle_message` instead to avoid this private call.

rasa_core/server.py
```diff
@@ -288,6 +290,50 @@ def parse(self, request, sender_id):
...
+            responses = processor.handle_message(UserMessage(message, out, sender_id))
+            request.setResponseCode(200)
+            return json.dumps(responses)
+
+        except Exception as e:
+            request.setResponseCode(500)
+            logger.error("Caught an exception during "
+                         "parse: {}".format(e), exc_info=1)
```
tmbo: not `parse` but `respond`

tests/test_agent.py
```diff
@@ -41,4 +41,5 @@ def test_agent_handle_message(default_agent):
     message = INTENT_MESSAGE_PREFIX + 'greet{"name":"Rasa"}'
     result = default_agent.handle_message(message,
                                           sender_id="test_agent_handle_message")
-    assert result == ["hey there Rasa!"]
+    assert result == [{'recipient_id': u'test_agent_handle_message',
```
tmbo: `u` should not be necessary, due to `from __future__ import unicode_literals`

tests/test_dispatcher.py
```diff
@@ -77,10 +82,14 @@ def test_dispatcher_utter_custom_message(default_dispatcher_collecting):
...
     collected = default_dispatcher_collecting.output_channel.messages
-    assert len(collected) == 6
-    assert collected[0][1] == "hey there : welcome"
-    assert collected[1][1] == "1: Btn1 (/btn1)"
-    assert collected[2][1] == "2: Btn2 (/btn2)"
-    assert collected[3][1] == "another title : another subtitle"
-    assert collected[4][1] == "1: Btn3 (/btn3)"
-    assert collected[5][1] == "2: Btn4 (/btn4)"
+    assert len(collected) == 2
+    assert collected[0]['text'] == "hey there : welcome"
```
tmbo: cool due to the dict the tests are way more readable :+1:

tests/test_trackers.py
```diff
@@ -105,8 +105,8 @@ def test_tracker_write_to_story(tmpdir, default_domain):
     recovered = trackers[0]
     assert len(recovered.events) == 8
     assert recovered.events[6].type_name == "slot"
-    assert recovered.events[6].key == "location"
-    assert recovered.events[6].value == "central"
+    assert recovered.events[6].key in {"location", "name"}
```
tmbo: why does this vary?

amn41: no idea - for some reason the order was one way in python 2.7 & 3.6, and another in 3.5. Will have another look before merging

rasa_core/server.py
amn41: that doesn't work because `handle_message` takes only text as input, we need a `UserMessage` so that we can specify the output channel & sender id. Happy to modify the handle_message method to take either though

tmbo: `handle_message` should already accept the output channel:

```
    def handle_message(
            self,
            text_message,  # type: Text
            message_preprocessor=None,  # type: Optional[Callable[[Text], Text]]
            output_channel=None,  # type: Optional[OutputChannel]
            sender_id=UserMessage.DEFAULT_SENDER_ID  # type: Optional[Text]
    ):
        # type: (...) -> Optional[List[Text]]
        """Handle a single message.
```

tests/test_trackers.py
tmbo: the test has been flaky on master as well and fails from time to time there as well.

rasa_core/server.py
amn41: oh nice!

amn41:
ok @tmbo back to you

PR #271, opened 2018-03-14
webhook was not set previously

tiwari85aman:
telegram integration was not working prior because the webhook required for the integration was not set previously.

Now, the telegram integration is working correctly as expected

- [ ] updated the changelog

### DIFF
--- a/rasa_core/channels/telegram.py
+++ b/rasa_core/channels/telegram.py
@@ -105,7 +105,8 @@ class TelegramInput(HttpInputComponent):
                 return "Webhook setup successful"
             else:
                 return "Invalid webhook"
-
+        set_webhook()
+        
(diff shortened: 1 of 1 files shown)


tiwari85aman:
@akelad is there anything else needed?

tmbo:
Given that the return value is ignored for the function, it would be good to add some logging. Do you mind adding a `logger.info` in the success case and a `logger.warn` in the failure case?

tiwari85aman:
@tmbo ok sure I will make those loggings. thanks for suggesting :+1:

tiwari85aman:
@tmbo sir, please verify the changes of logger updat.

tiwari85aman:
@tmbo is there something else required, apart from logger update?

tmbo:
No looking good, thanks for the fix :+1:

PR #301, opened 2018-03-22
Simpler generator

### LINKED ISSUE #286: Split action names into labels to allow for generalisation experiments
Right now the label of an action is directly predicted. Nevertheless, alot of the actions shore a common name pattern that yields useful information for subclustering, e.g. `action_request_date`, `action_confirm_date`, `action_invalid_date` - all of these somehow belong to each other.

If we split the action names on special characters, e.g. `_` we can use the resulting label set as the target for the classification instead of the whole label. this way, the prediction of the `date` label would be shared for the above three examples.

### LINKED ISSUE #387: dump flattened stories into trained model dir

### LINKED ISSUE #449: where to find log files?
**Rasa Core version**:
Master (commit 241979ca4fbe22083aec7b1b6c37810cc0ce2e6c)

**Python version**: 
3.5.2
linux
**Issue**:
Hi, I run the bot in a terminal, using simply:

```
def run(serve_forever=True):

    interpreter = RasaNLUInterpreter("a_path")
    agent = Agent.load("a_different_path", interpreter=interpreter)

    if serve_forever:
        agent.handle_channel(ConsoleInputChannel())
    return agent
```
It would be great if I could specify the location of a logfile, where for example all the user input and interpreter output would be stored. If such a file is already being created, it is indeed very well hidden, so a pointer to its default location would be much appreciated :) 

Without such a file, the only way to debug a weird conversation is to exit the bot and launch the online training, trying to reproduce the same conversation flow...

Thanks for your help,

Andrea.

amn41:
**Proposed changes**:
- keras policy now fits entire dialogue sequences simultaneously or slices of history to a single action, depending on the Featurizer provided
- let the `Featurizer` class do the featurization
- `generate()` just returns a list of finished trackers
- `max_history` now an attribute of the featurizer
- separate method for featurizing labels in Featurizer class

- fixes #286
- fixes #387
- fixes #449

- [ ] updated the changelog


Ghostvv:
Could you please add a bit more details on `what's going wrong with keras policy training`?

amn41:
sure - when I train for a large number of epochs, accuracy quickly goes up to e.g. 0.9, then steadily drops to ~0.6 while loss continues to decrease. 

Also my longest dialogue is 9 actions, I train with max_history = 10 vs 13. With 13 the same behaviour occurs but more slowly. These extra steps should be ignored anyway and not affect the optimisation at all

rasa_core/featurizers.py
```diff
@@ -81,6 +83,62 @@ class BinaryFeaturizer(Featurizer):
...
+        max_len = self.max_history
+        trackers_actions = []
+        for tracker in trackers:
+            actions = []
+            for event in tracker._applied_events():
+                 if isinstance(event, ActionExecuted):
+                     actions.append(event.action_name)
+            trackers_actions.append(actions)
+        # slice in case longer than max_len
+        trackers_actions = trackers_actions[-self.max_history:]
```
Ghostvv: this is wrong, because `len(trakers) = len(trackers_actions) = num stories` and `max_history` limits the dialogue length

Ghostvv:
Right now, policy receives training_data for training, but tracker for prediction. Let's change that.

We are going to pass training trackers to train method of the policy. The train method will call featurizer as it is done in predict method.

tmbo:
Status: @Ghostvv will push his local changes and we'll do a first review of this tomorrow.

Ghostvv:
@tmbo I pushed the last changes. On my computer, training and prediction of `memoization` and `keras` policies work

Ghostvv:
@tmbo @amn41 I updated all main methods to the new system

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+            return np.ones(num_features, dtype=int) * -1
+
+        used_features = np.zeros(num_features, dtype=int)
+        for feature_name, prob in active_features.items():
+
+            if feature_name in user_labels:
+                if 'prev_action_listen' not in active_features:
+                    # we predict next action from bot action
+                    # feature_name = 'intent_listen'
+                    used_features[:len(self.user_vocab)] = 0
```
tmbo: this is a change right? so up until now we used the intent of the previous message for all action predections not only for the first one after a user message, right?

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+                else:
+                    for t in feature_name.split(self.split_symbol):
+                        used_features[self.user_vocab[t]] += 1
+
+            elif feature_name[len('prev_'):] in bot_labels:
```
tmbo: I think we should change the structure of the collected features and making them objectes (containing name, type and value) - but let's not do it as part of this pr.

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+                for t in feature_name[len('prev_'):].split(self.split_symbol):
+                    used_features[len(self.user_vocab) +
+                                  self.bot_vocab[t]] += 1
+
+            elif feature_name in self.additional_features:
+                if prob > 0:
```
tmbo: why would it be an active feature with `prob == 0.0`?

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+    def _pad_states(self, states):
+        return states
+
+    def _featurize_states(self, trackers_as_states, domain):
+        """Create X"""
+        features = []
+        true_lengths = []
+
+        for tracker_states in trackers_as_states:
+            if len(trackers_as_states) > 1:
```
tmbo: why do we only pad in this case?

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+        X = np.array(features)
+
+        return X, true_lengths
+
+    def _featurize_labels(self, trackers_as_actions, domain):
+        """Create y"""
+
+        labels = []
+        for tracker_actions in trackers_as_actions:
+            if isinstance(tracker_actions, list):
```
tmbo: why is it sometimes a list?

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+
+            trackers_as_states.append(states[:-1])
+            trackers_as_actions.append(actions)
+
+        if self.max_len is None:
+            self._calculate_max_len(trackers_as_actions)
+
+        # TODO do we need that?
+        metadata = {"events": events_metadata,
+                    "trackers": trackers}
```
tmbo: i think the trackers were needed for the visualisation (but we don't need that anymore if we create the trackers before the featurization)

events metadata is used for warnings when using the bot (e.g. if a slot is set after an action, that has never been set in the training data after this specific action)

rasa_core/featurizers.py
```diff
@@ -6,83 +6,56 @@
...
+if typing.TYPE_CHECKING:
+    from rasa_core.trackers import DialogueStateTracker
+    from rasa_core.domain import Domain
 
-class Featurizer(object):
-    """Transform the conversations state into machine learning formats.
 
-    Featurizer decides how the bot will transform the conversation state to a
-    format which a classifier can read."""
+class FeaturizeMechanism(object):
```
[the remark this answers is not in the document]
tmbo: reading through it I think the difference between a featurizer and a featurizemechanism is that the featurizer featurizes a **whole tracker** wheras the mechanism featurizes a **single state** right? so lets name it that way

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+
+
+class FullDialogueFeaturizer(Featurizer):
+    def __init__(self, featurize_mechanism):
+        # type: (FeaturizeMechanism) -> None
+        super(FullDialogueFeaturizer, self).__init__(featurize_mechanism)
+
+        self.max_len = None
+
+    def _calculate_max_len(self, as_states):
```
tmbo: needs short description

tmbo: what does `as_states` mean?

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+        self.max_len = 0
+        for states in as_states:
+            self.max_len = max(self.max_len, len(states))
```
tmbo: lets do 
```
if not as_states:
   return 0
else:
   return max([len(states) for states in as_states])

File: rasa_core/featurizers.py [hunk 11]
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+        logger.info("The longest dialogue has {} actions."
+                    "".format(self.max_len))
+
+    def _pad_states(self, states):
```
tmbo: so padding moved from the beginning to the end? i think we discussed that, just want to make sure

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+            for event in tracker._applied_events():
+                if isinstance(event, ActionExecuted):
+                    if not event.unpredictable:
+                        # only actions which can be predicted at a stories start
+                        actions.append(event.action_name)
+                    else:
+                        # unpredictable actions can be only the first in the story
+                        delete_first_state = True
+                else:
+                    action_name = tracker.latest_action_name
```
tmbo: I don't think this is correct: the `tracker.latest_action_name` will always be the same while we iterate through the events of the tracker. so the event (e.g. slot set) will always be counted towards the last **action of the whole tracker** but what we need is the previous action within the tracker. do you know what I mean?

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+            delete_first_state = False
+            actions = []
+            for event in tracker._applied_events():
+                if isinstance(event, ActionExecuted):
+                    if not event.unpredictable:
+                        # only actions which can be predicted at a stories start
+                        actions.append(event.action_name)
+                    else:
+                        # unpredictable actions can be only the first in the story
+                        delete_first_state = True
```
tmbo: let's raise an exception if `delete_first_state ` is already set to `True` (meaning that there are two actions that are unpredictable, which shouldn't really be the case)

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+                else:
+                    action_name = tracker.latest_action_name
+                    events_metadata[action_name].add(event)
+
+            if delete_first_state:
+                states = states[1:]
+
+            trackers_as_states.append(states[:-1])
```
tmbo: whats the `-1` for?

rasa_core/featurizers.py
tmbo: let's return the max len instead of directly setting it to the variable.

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+                if isinstance(event, ActionExecuted):
+                    if not event.unpredictable:
+                        # only actions which can be predicted at a stories start
+                        sliced_states = domain.slice_feature_history(
+                            states[:idx + 1], self.max_history)
+                        trackers_as_states.append(sliced_states)
+                        trackers_as_actions.append(event.action_name)
+                    idx += 1
+                else:
+                    action_name = tracker.latest_action_name
```
tmbo: same as in other featurizer

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+        trackers_as_states = []
+        trackers_as_actions = []
+        events_metadata = defaultdict(set)
+
+        for tracker in trackers:
+            states = domain.states_for_tracker_history(tracker)
+
+            delete_first_state = False
+            actions = []
+            for event in tracker._applied_events():
```
tmbo: I don't think we can only go through the applied events. e.g. consider:

```
## my nasty story
* greet
   - greet
   - undo
```
in this case the greet action will be reverted. while training the model we never generate the training sample to predict the action greet.

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+        # type: (...) -> Tuple[List[List[Dict]], List[List[Dict]], Dict]
+
+        trackers_as_states = []
+        trackers_as_actions = []
+        events_metadata = defaultdict(set)
+
+        for tracker in trackers:
+            states = domain.states_for_tracker_history(tracker)
+            idx = 0
+            for event in tracker._applied_events():
```
tmbo: same as in other featurizer

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+            trackers_as_states,  # type: List[List[Dict[Text, float]]]
+            trackers_as_actions,  # type: List[List[Dict[Text, float]]]
+    ):
+        # type: (...) -> Tuple[List[List[Dict]], List[List[Dict]]]
+        """Removes states that create equal featurizations.
+
+        From multiple states that create equal featurizations
+        we only need to keep one."""
+
+        hashed_featurizations = []
```
tmbo: should be a set

rasa_core/agent.py
```diff
@@ -40,7 +40,7 @@ def __init__(
             tracker_store=None  # type: Optional[TrackerStore]
     ):
         self.domain = self._create_domain(domain)
-        self.featurizer = self._create_featurizer(featurizer)
```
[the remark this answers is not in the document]
Ghostvv: ok

rasa_core/agent.py
```diff
@@ -155,6 +155,7 @@ def train_online(self,
                      resource_name=None,  # type: Optional[Text]
                      input_channel=None,  # type: Optional[InputChannel]
                      model_path=None,  # type: Optional[Text]
+                     max_history=2,  # type: int
```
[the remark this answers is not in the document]
Ghostvv: Or 3? https://github.com/RasaHQ/rasa_core/blob/77a479567c298e12bc6a0245dd17324ff535b3a3/rasa_core/policies/online_policy_trainer.py#L34

rasa_core/domain.py
```diff
@@ -179,27 +178,15 @@ def slice_feature_history(self,
...
-        encoded_features = [featurizer.encode(f, self.input_feature_map)
-                            for f in state_features]
-        return np.vstack(encoded_features)
+        return state_features
 
-    def features_for_tracker_history(self, tracker):
-        """Array of features for each state of the trackers history."""
+    def states_for_tracker_history(self, tracker):
+        # type: (DialogueStateTracker) -> List[Dict[Text, float]]
+        """Array of states for each state of the trackers history."""
```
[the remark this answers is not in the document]
Ghostvv: yes, by `state` mean `dictionary of features`, so it will return list of dictionaries of features

rasa_core/featurizers.py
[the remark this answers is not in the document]
Ghostvv: right

rasa_core/featurizers.py
Ghostvv: because during training we have the last state in tracker after last `action_listen` which is `[{...<slots>..., 'intent_<last>':1.0, 'prev_action_listen':1}]`

rasa_core/featurizers.py
Ghostvv: in generator as well then:
https://github.com/RasaHQ/rasa_core/blob/031df4c36b48fc26e9bc2765c1394ad38ff2c50f/rasa_core/training/generator.py#L58

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+        user_labels, bot_labels, other_labels = self.labels
+
+        self.bot_actions = self._create_label_dict(bot_labels)
+
+        if not self.share_vocab:
+            self.bot_vocab = self._create_label_token_dict(bot_labels,
+                                                           self.split_symbol)
+            self.user_vocab = self._create_label_token_dict(user_labels,
+                                                            self.split_symbol)
+        else:
```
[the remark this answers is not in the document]
Ghostvv: what do you mean? `share_vocab` is a flag if to use common dict for both user and bot or two different dicts.

rasa_core/featurizers.py
```diff
@@ -97,7 +70,7 @@ def encode(self, active_features, input_feature_map):
         padding vectors are specified by a `None` or `[None]`
         value for active_features."""
 
-        num_features = len(input_feature_map.keys())
+        num_features = domain.num_features
```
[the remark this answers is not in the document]
Ghostvv: this `num_features` is not a number of numeric features, but rather the number of string keys like `intent_greet`, `prev_action_listen` etc. Then we need to consider refactoring domain, if want to change that or come up with a better name

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+
+    def encode(self, active_features, domain):
+
+        user_labels, bot_labels, other_labels = self.labels
+
+        num_features = (len(self.user_vocab) +
+                        len(self.bot_vocab) +
+                        len(self.additional_features))
+
+        if active_features is None or None in active_features:
```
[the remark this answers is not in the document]
Ghostvv: I do not know, I copied it from the old code:
https://github.com/RasaHQ/rasa_core/blob/c17193d46311cbbbb67fd35295e964ed880463ef/rasa_core/featurizers.py#L101

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+            return np.ones(num_features, dtype=int) * -1
+
+        used_features = np.zeros(num_features, dtype=int)
+        for feature_name, prob in active_features.items():
+
+            if feature_name in user_labels:
+                if 'prev_action_listen' not in active_features:
```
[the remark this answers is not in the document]
Ghostvv: do we have this constant somewhere, where this feature string was created?

rasa_core/featurizers.py
Ghostvv: yes, I copied it from another new policy

rasa_core/featurizers.py
Ghostvv: yes, good idea, let's do it together with refactoring domain

rasa_core/featurizers.py
Ghostvv: because the slots are always present in `states`=`list of feature dicts` with 0.0 as values if they are not set

rasa_core/featurizers.py
Ghostvv: because during prediction `len(trackers_as_states) == 1` and since we use `dynamic_rnn` in keras, we do not need to pad

rasa_core/featurizers.py
Ghostvv: for `FullDialogueFeaturizer` it is a list, for `MaxHistoryFeaturizer` it is not

rasa_core/featurizers.py
Ghostvv: I would be glad to get rid of that, it makes a lot of functions ugly. If we need it, it is maybe make sense to add such method in tracker

rasa_core/featurizers.py
Ghostvv: yes for Full dialogue featurizer - yes, for Max history it is the old style and is still done by domain (do we want to change it now?)

rasa_core/featurizers.py
Ghostvv: that's why i got strange results :)
But if we ditch metadata, we do not need that here

rasa_core/featurizers.py
Ghostvv: It would be a good idea to change that, because it makes nearly impossible to figure out what's going on when you print `states`

rasa_core/agent.py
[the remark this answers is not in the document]
Ghostvv: plus it supports back compatibility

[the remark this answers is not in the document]
Ghostvv: done

rasa_core/featurizers.py
```diff
@@ -163,18 +136,425 @@ def encode(self, active_features, input_feature_map):
...
+            elif feature_name.startswith('prev_'):
+                bot_labels.append(feature_name[len('prev_'):])
+            else:
+                other_labels.append(feature_name)
+
+        return user_labels, bot_labels, other_labels
+
+    @staticmethod
+    def _create_label_dict(labels):
+        """Create label dictionary"""
```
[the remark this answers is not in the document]
Ghostvv: let's deal with descriptions later, after we refactor the code most of the functions repeat domain functionality

rasa_core/featurizers.py
Ghostvv: I think the generator should return events metadata

rasa_core/featurizers.py
Ghostvv: I think the generator should return events metadata, and then we can pass it to ensemble for persist

rasa_core/featurizers.py
Ghostvv: yes, but how would you change that?

rasa_core/featurizers.py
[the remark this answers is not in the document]
Ghostvv: use case is if want to merge embeddings for inputs and bot actions to infer the context directly, not sure if it is possible though

rasa_core/featurizers.py
Ghostvv: we can go around here, by not adding the first action listen, instead of forgetting the last one

Ghostvv: It should be fine because during training we do not see explicitly `ActionReverted` or `UserUtteredReverted` because they are not part of the stories, so during training `_applied_events()=events` except for double action_listen due to augmentation.

During prediction phase, we go step by step so we should be able to predict `greet` and `undo` in this case. After undo we'll try to predict as if they never happen

rasa_core/policies/augmented_memoization.py
```diff
@@ -0,0 +1,106 @@
...
+                        # the last historic_events repeat the tracker
+                        # so we delete them
+                        del historic_events[-1]
+                        break
+                    historic_events.append(historic_events[i - 1][::-1])
+
+            historic_events.reverse()
+
+            for events in historic_events:
+                mcfly_tracker = tracker._init_copy()
```
tmbo: lets get rid of `_` here as well on `init_copy`

rasa_core/policies/embedding_policy.py
```diff
@@ -0,0 +1,830 @@
+from __future__ import absolute_import
```
tmbo: TODO: continue review

rasa_core/policies/ensemble.py
```diff
@@ -10,84 +10,89 @@
...
         logger.debug("Predicted next action '{}' with prob {:.2f}.".format(
                 domain.action_for_index(max_index).name(),
                 probabilities[max_index]))
         return max_index
 
-    def probabilities_using_best_policy(self, tracker, domain):
-        raise NotImplementedError
+    def _max_histories(self):
+        # type: () -> List[Optional[int]]
+        """Return max history, only works if the ensemble is already trained."""
```
tmbo: can we raise an exception if it is not?

rasa_core/policies/ensemble.py
```diff
@@ -10,84 +10,89 @@
...
+                max_histories.append(None)
+        return max_histories
 
     @staticmethod
     def _create_action_fingerprints(training_events):
         """Fingerprint each action using the events it created during train.
 
         This allows us to emit warnings when the model is used
         if an action does things it hasn't done during training."""
+        if training_events:
```
tmbo: why would they not be present?

rasa_core/policies/ensemble.py
```diff
@@ -10,84 +10,89 @@
...
+            action_fingerprints = {}
+            for k, vs in training_events.items():
+                slots = list({v.key for v in vs if isinstance(v, SlotSet)})
+                action_fingerprints[k] = {"slots": slots}
+            return action_fingerprints
+        else:
```
tmbo: let's avoid the nesting:
```python
def _create_action_fingerprints(training_events):
  if not training_events:
    return None
  
  action_fingerprints = {}
  ...
  return action_fingerprints
```

rasa_core/policies/ensemble.py
```diff
@@ -113,10 +117,13 @@ def persist(self, path):
...
         """Persists the policy to storage."""
 
-        self._persist_metadata(path, self.max_history())
+        self._persist_metadata(path, self._max_histories())
 
-        for policy in self.policies:
-            policy.persist(path)
+        for i, policy in enumerate(self.policies):
+            # TODO better way then many folders?
+            # TODO delete old files from these folders
```
tmbo: is there a reason not to use one folder? if there is an issue with that multiple folders are fine.

rasa_core/policies/ensemble.py
```diff
@@ -144,9 +152,12 @@ def load(cls, path, featurizer):
 
 class SimplePolicyEnsemble(PolicyEnsemble):
     def __init__(self, policies, known_slot_events=None):
+        # TODO is known_slot_events is the same as action_fingerprints?
```
tmbo: yes they are the same

examples/restaurantbot/bot.py
```diff
@@ -67,12 +67,12 @@ def train_dialogue(domain_file="restaurant_domain.yml",
...
 def train_nlu():
-    from rasa_nlu.converters import load_data
-    from rasa_nlu.config import RasaNLUConfig
+    from rasa_nlu.training_data import load_data
+    from rasa_nlu import config
     from rasa_nlu.model import Trainer
 
     training_data = load_data('data/franken_data.json')
-    trainer = Trainer(RasaNLUConfig("nlu_model_config.json"))
+    trainer = Trainer(config.load("nlu_model_config.json"))
```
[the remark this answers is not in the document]
Ghostvv: ye, I already merged master

rasa_core/featurizers.py
Ghostvv: I append copy of the tracker to the list of trackers here:
https://github.com/RasaHQ/rasa_core/blob/1f991980cbb9387e9a5da9bcc69777c2fabb3264/rasa_core/training/generator.py#L247

examples/restaurantbot/policy.py
```diff
@@ -11,19 +11,34 @@
...
         n_hidden = 32  # size of hidden layer in LSTM
         # Build Model
-        batch_shape = (None, max_history_len, num_features)
-
         model = Sequential()
-        model.add(Masking(-1, batch_input_shape=batch_shape))
-        model.add(LSTM(n_hidden, batch_input_shape=batch_shape))
-        model.add(Dense(input_dim=n_hidden, output_dim=num_actions))
+
+        if len(output_shape) == 1:
```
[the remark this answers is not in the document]
Ghostvv: totally agree

rasa_core/agent.py
```diff
@@ -155,6 +151,7 @@ def train_online(self,
                      resource_name=None,  # type: Optional[Text]
                      input_channel=None,  # type: Optional[InputChannel]
                      model_path=None,  # type: Optional[Text]
+                     max_history=3,  # type: int
```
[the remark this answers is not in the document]
Ghostvv: this is the history of outputting previous actions as well

rasa_core/featurizers.py
```diff
@@ -161,20 +157,443 @@ def encode(self, active_features, input_feature_map):
...
+        if states is None or None in states:
+            return np.ones(self.num_features, dtype=int) * -1
+
+        used_features = np.zeros(self.num_features, dtype=int)
+        for state_name, prob in states.items():
+
+            if state_name in self.user_labels:
+                if PREV_PREFIX + ACTION_LISTEN_NAME not in states:
+                    # we predict next action from bot action
+                    # TODO do we need state_name = 'intent_listen' ?
```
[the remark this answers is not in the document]
Ghostvv: since I deleted previous intent from feature list, if user did not input anything, I was thinking whether it'd make sense to create `intent_listen` instead of zero feature vector

rasa_core/policies/augmented_memoization.py
```diff
@@ -0,0 +1,106 @@
...
+logger = logging.getLogger(__name__)
+
+if typing.TYPE_CHECKING:
+    from rasa_core.trackers import DialogueStateTracker
+    from rasa_core.domain import Domain
+
+ENABLE_FEATURE_STRING_COMPRESSION = True
+
+
+class AugmentedMemoizationPolicy(MemoizationPolicy):
```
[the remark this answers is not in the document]
Ghostvv: yes, because Memoization sets the probability to `1`, the fact that it does additional staff might create confusion, so I split them. This AugmentedMemoizationPolicy is more elaborate version of scoring policy

rasa_core/policies/augmented_memoization.py
```diff
@@ -0,0 +1,106 @@
...
+            augmented_states[i] = None
+            augmented.append(list(augmented_states))
+        return augmented
+
+    def _back_to_the_future(self, tracker):
+        mcfly_trackers = []
+        if self.max_history > 1:
+            historic_events = [[]]
+            i = 0
+            e_i_last = len(tracker._applied_events()) - 1
```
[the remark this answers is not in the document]
Ghostvv: saw in your PR

rasa_core/policies/augmented_memoization.py
```diff
@@ -0,0 +1,106 @@
...
+            for e_i, event in enumerate(
+                    reversed(tracker._applied_events())):
```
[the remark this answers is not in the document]
Ghostvv: definitely

rasa_core/policies/augmented_memoization.py
```diff
@@ -0,0 +1,106 @@
...
+                historic_events[i].append(event)
+
+                if isinstance(event, ActionExecuted):
+                    historic_events[i].reverse()
+                    i += 1
+                    if i == self.max_history - 1:
+                        # we need i to be one less than max_history
+                        # not to recall again with the same features
```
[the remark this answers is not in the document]
Ghostvv: because when `i = max_history`, we'll create exactly the same tracker, as it was passed, so there is no need to try to recall it again

rasa_core/policies/ensemble.py
Ghostvv: if it is not = if the ensemble is trained?

rasa_core/policies/ensemble.py
Ghostvv: I guess, if it was not trained

rasa_core/policies/ensemble.py
Ghostvv: this is the fastest solution, i came with to avoid renaming feauturizers, because we have multiple ones now

Ghostvv: then we would need to clean the folder before writing, because if you changed policy, old files will be there, and it looks confusing

rasa_core/policies/ensemble.py
Ghostvv: let's rename it here maybe?

rasa_core/policies/ensemble.py
Ghostvv: that's actually wrong description, it would work before training as well

rasa_core/policies/ensemble.py
Ghostvv: also, if you have two keras policies with different featurizer, I think it is clearer to keep them in different folders

rasa_core/policies/policy.py
```diff
@@ -4,74 +4,131 @@
 from __future__ import unicode_literals
 
 import logging
+import typing
 
-import numpy as np
+from inspect import signature
```
akelad: @Ghostvv this is incompatible with python 2.7

rasa_core/featurizers.py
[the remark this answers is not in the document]
tmbo: ah ok, so when the conversation is started and the state list is still completely empty?

rasa_core/policies/policy.py
akelad: i've fixed it with using `valid_keys = getargspec(func).args` instead, as far as i know it does the same thing , it just returns a list instead of a dict

rasa_core/policies/embedding_policy.py
```diff
@@ -0,0 +1,848 @@
...
+
+        "droprate_a": 0.1,
+        "droprate_b": 0.1,
+        "droprate_c": 0.3,
+        "droprate_rnn": 0.1,
+        "droprate_out": 0.1,
+    }
+
+    def _load_nn_architecture_params(self):
+        self.num_hidden_layers_a = self.config['num_hidden_layers_a']
```
tmbo: I am not quite sure, but what is the benefit of setting all of these as single attributes?

rasa_core/policies/embedding_policy.py
```diff
@@ -0,0 +1,848 @@
...
+        self.encoded_all_actions = self._create_encoded_actions(domain)
+
+        # check if number of negatives is less than number of actions
+        logger.debug("Check if num_neg {} is smaller "
+                     "than number of actions {}, "
+                     "else set num_neg to the number of actions - 1"
+                     "".format(self.num_neg, domain.num_actions))
+        self.num_neg = min(self.num_neg, domain.num_actions - 1)
+
+        X, Y, extras, actions_for_X = self._prepare_data_for_training(
```
tmbo: there is a lot of data munging going on - are there any tests to make sure the representation during training is the same as during prediction?

rasa_core/policies/embedding_policy.py
```diff
@@ -0,0 +1,848 @@
...
+        # type: (DialogueStateTracker, Domain) -> List[float]
+        """Predicts the next action the bot should take
+        after seeing the tracker.
+
+        Returns the list of probabilities for the next actions"""
+        if self.session is None:
+            logger.error("There is no trained tf.session: "
+                         "component is either not trained or "
+                         "didn't receive enough training data")
+            return []
```
tmbo: this should still have num actions shape I guess (or be `None` - but not sure this will be handled properly in ensemble)

rasa_core/policies/embedding_policy.py
```diff
@@ -0,0 +1,848 @@
...
+        if self.session is None:
+            warnings.warn("Persist called without a trained model present. "
+                          "Nothing to persist then!")
+            return
+
+        file_name = 'tensorflow_embedding.ckpt'
+        checkpoint = os.path.join(path, file_name)
+
+        try:
+            os.makedirs(os.path.dirname(checkpoint))
```
tmbo: if seen this in the code base before, but given that the try catch is quite large, I'd vote for:
```
if not os.path.exists(path):
  os.makedirs(path)
```

rasa_core/policies/ensemble.py
tmbo: ok sounds good

rasa_core/policies/keras_policy.py
```diff
@@ -122,6 +202,10 @@ def _persist_configuration(self, config_file):
         utils.dump_obj_as_json_to_file(config_file, model_config)
 
     def persist(self, path):
+        # type: (Text) -> None
+
+        super(KerasPolicy, self).persist(path)
```
tmbo: i am not so sure about this style wise: in persist we use super implementation (but in load we do not).

rather put the `self.featurizer.persist(path)` into `if self.model` branch, as there must be a featurizer present right?

rasa_core/policies/online_policy_trainer.py
```diff
@@ -6,6 +6,9 @@
 import copy
 import logging
 import os
+import typing
```
tmbo: TODO: continue reviewing.

rasa_core/featurizers.py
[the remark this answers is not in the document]
Ghostvv: yes, but I think that is not needed, I'll leave this TODO for now as a reminder, we'll remove it before merge

rasa_core/policies/ensemble.py
Ghostvv: I think agent should clear model folder before writing into it, could you please recommend a good way to do it?

rasa_core/policies/embedding_policy.py
Ghostvv: I call the same function to create X during training and prediction

rasa_core/policies/embedding_policy.py
```diff
@@ -0,0 +1,848 @@
...
+    # data helpers:
+    def _create_all_Y_d(self, dialogue_len):
+        # stack encoded_all_intents on top of each other
+        # to create candidates for training examples
+        # to calculate training accuracy
+        all_Y_d = np.stack([self.encoded_all_actions
+                            for _ in range(dialogue_len)])
+        return all_Y_d
+
+    def _create_X_extras(self, data_X):
```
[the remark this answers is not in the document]
Ghostvv: extras are effectively slots, I renamed them to slots

rasa_core/policies/embedding_policy.py
Ghostvv: yes, list of zeros, i think

rasa_core/policies/embedding_policy.py
Ghostvv: it is the code from https://github.com/RasaHQ/rasa_core/blob/811d42b1d29c40a42d7dee345572ce73330b8360/rasa_core/utils.py#L133
I will call this function here

rasa_core/policies/ensemble.py
tmbo: yes that is a good idea. 

`shutil.rmtree(model_dir)` but there should be a check if that folder really contains a model (e.g. check if `policy_metadata.json` is present or something like that)

rasa_core/policies/embedding_policy.py
tmbo: even better :+1:

rasa_core/policies/embedding_policy.py
tmbo: no I mean in general, store the whole config `self.config = config` and where needed use `self.config['num_hidden_layers_a']` instead of `self.num_hidden_layers_a` (avoiding all the attributes)

tmbo: I don't have a strong opinion about this, we can also keep it that way (but then `config` shouldn't be an object property as it will duplicate the values and they will run out of sync).

rasa_core/policies/embedding_policy.py
Ghostvv: but yeah, I guess it is still good, to create a test for it

rasa_core/policies/embedding_policy.py
```diff
@@ -0,0 +1,848 @@
...
+        self.graph = graph
+        self.a_in = intent_placeholder
+        self.b_in = action_placeholder
+        self.c_in = extras_placeholder
+        self.sim_op = similarity_op
+        # for continue training
+        self.train_op = None
+        self.is_training = None
+
+    def _create_encoded_actions(self, domain):
```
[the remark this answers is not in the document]
Ghostvv: I moved it to label state featurizer

rasa_core/policies/sklearn_policy.py
```diff
@@ -139,44 +154,57 @@ def train(self, training_data, domain, **kwargs):
...
-        X, y = self._extract_training_data(training_data)
-        Xt, yt = self._preprocess_data(X, y)
-        if not hasattr(self.model, 'partial_fit'):
-            raise TypeError("Continuing training is only possible with "
-                            "sklearn models that support 'partial_fit'.")
-        self.model.partial_fit(Xt, yt)
+    def _postprocess_prediction(self, y_proba, domain):
+        yp = y_proba[0].tolist()
+
+        # Some classes might not be part of the training labels. Since
```
tmbo: nice :+1:

rasa_core/policies/sklearn_policy.py
```diff
@@ -139,44 +154,57 @@ def train(self, training_data, domain, **kwargs):
...
+        featurizer = TrackerFeaturizer.load(path)
+        assert isinstance(featurizer, MaxHistoryTrackerFeaturizer), \
+            ("Loaded featurizer of type {}, should be "
+             "MaxHistoryTrackerFeaturizer.".format(type(featurizer).__name__))
+
+        policy = cls(featurizer=featurizer)
+
         with open(filename, 'rb') as f:
             state = pickle.load(f)
+        vars(policy).update(state)
```
tmbo: does this work? I think that dict is write protected for new style classes? but yeah you didn't write that - just noticed

rasa_core/training/generator.py
```diff
@@ -315,25 +237,29 @@ def _process_step(self, step, incoming_trackers):
         # need to copy the tracker as multiple story steps
         # might start with the same checkpoint and all of them
         # will use the same set of incoming trackers
-        trackers = [tracker.create_copy() for tracker in
+        trackers = [deepcopy(tracker) for tracker in
```
tmbo: please avoid importing functions from somewhere, better call them on the package (e.g. `copy.deepcopy(...)`) as this makes clear where that function is coming from.

rasa_core/policies/sklearn_policy.py
Ghostvv: that's i do not know, we need to check

rasa_core/policies/embedding_policy.py
Ghostvv: tbh, I just copied it from nlu embedding classifier

rasa_core/policies/sklearn_policy.py
tmbo: don't worry about it now, it's nothing this PR changed.

rasa_core/policies/embedding_policy.py
```diff
@@ -0,0 +1,1022 @@
...
+        train_acc = np.sum((np.argmax(_sim, -1) ==
+                            actions_for_X[ids]) * _mask)
+        train_acc /= np.sum(_mask)
+        logger.info("epoch {} / {}: loss={:.3f}, accuracy={:.3f}"
+                    "".format((ep + 1), self.epochs,
+                              ep_loss, train_acc))
+
+        t = 9
+        for i, idx in enumerate(actions_for_X[ids[0], :t+1]):
+            print("{:.3f} -- {}".format(_x[0, t, i], self.domain.actions[idx].name()))
```
tmbo: @Ghostvv can we delete this and get rid of the domain here?

rasa_core/featurizers.py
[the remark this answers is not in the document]
tmbo: @Ghostvv so should we remove this, wdyt?

rasa_core/policies/embedding_policy.py
```diff
@@ -0,0 +1,1022 @@
...
+        shape_b = tf.shape(b)
+        b = tf.layers.dropout(b, rate=self.droprate['b'],
+                              training=self.is_training,
+                              noise_shape=[shape_b[0],
+                                           shape_b[1],
+                                           1, shape_b[-1]])
+        emb_act = self._create_embed(b, name=name_b)
+
+        c = c_in
+        # TODO do we need hidden layers for slots?
```
tmbo: @Ghostvv do we?

rasa_core/policies/embedding_policy.py
Ghostvv: yes i will delete it soon, I still use it for testing though

rasa_core/policies/embedding_policy.py
Ghostvv: i don't think so, i set it as a reminder for myself

tmbo:
Are there any other todos in your mind @Ghostvv ? I went through the code ones and they look good.

tmbo:
The policy isn't covered in the documentation yet is it? Should get a section and a short explanation.

Ghostvv:
No, I didn't write anything about it

tmbo:
Mind writing something, I mean it is only of so much use if nobody knows that it exists why it is cool and when it should be used.

amn41:
I think for now a small amount of documentation will suffice (just to say it exists) - the "how and when to use it" will probably get wrapped into docs on generalisation

Ghostvv:
I was planning to do docs right before merging, to not waste time rewriting them every time I change smth

Ghostvv:
@tmbo we also need new docs for featurization

Ghostvv:
@tmbo what about checking in the train method of `keras_policy` if `model==None` then call `self.model_architecture(..)` else use passed `model`?

tmbo:
What's the use case for passing in a model through the train function?

Ghostvv:
no, not through train function. Right now, in principle fresh `keras_policy` can be initialized with some keras model (we use it for persistence):
https://github.com/RasaHQ/rasa_core/blob/5616d3332ad22c5a29fb08b87e1dca3cfd101343/rasa_core/policies/keras_policy.py#L37

Why don't we check in `train` if it `is not None`?

tmbo:
Sure, feel free to change that.

docs/featurisation.rst
```diff
@@ -3,81 +3,96 @@
...
+
+        The target labels correspond to actions taken by the bot.
+        To convert the features into vector format, there are different
+        feaurizers available:
+
+        - ``BinarySingleStateFeaturizer`` creates a binary one-hot encoding:
+            The vectors ``X, y`` indicate a presence of a certain intent,
+            entity, previous action or slot e.g. ``[0 0 1 0 0 1 ...]``.
+
+        - ``LabelTokenizerSingleStateFeaturizer`` creates an embedding of the feature label:
```
Ghostvv: @tmbo I think using `embedding` here might be confusing

PR #320, opened 2018-03-29
Update connectors.rst

btotharye:
adding in section for slack to use ngrok will see how this looks and update the other sections as well.

**Proposed changes**:
- ...

- [ ] updated the changelog


akelad:
@btotharye looks good so far :) I'd actually maybe make the general instructions for setting up a webhook with ngrok a seperate section at the end, and then link to it from each section, adding any extra info necessary for the individual channels. e.g. with Facebook you need a static URL, because you can only set the webhook once

tmbo:
@btotharye do you have some time to look into the comment @akelad made? Would be great to pick up her idea and merge this afterwards.

btotharye:
Sure when I get back tomorrow I can knock it out

btotharye:
Sorry been travelling will fix it this week

btotharye:
ok I'm working on this today and will get it updated, we just want the ngrok section right @akelad

btotharye:
@akelad let me know what you think about what I have here so far

docs/connectors.rst
```diff
@@ -80,6 +80,24 @@ facebook developer portal.
...
+         permissions (for basic functionality you should subscribe to the ``message.channel``,
+         ``message.groups``, ``message.im`` and ``message.mpim`` events)
+      3. The ``slack_channel`` is the target your bot posts to. This can be a channel,
+         an app or an individual person
+      4. Use the entry for ``Bot User OAuth Access Token`` in the "OAuth & Permissions" tab
+         as your ``slack_token``
+
+
+   For more detailed steps, visit the
+   `slack api docs <https://api.slack.com/incoming-webhooks>`_.
```
akelad: can we actually just move this note back to where it was? looks a bit strange at the start of the section imo.

docs/connectors.rst
```diff
@@ -89,6 +107,14 @@ If you want to connect to the slack input channel using the run script, e.g. usi
...
       --port 5002 --connector slack --credentials slack_credentials.yml
 
+Setting Up Webhook
+^^^^^^^^^^^^^^^^^^
+In order to use this with slack you need a external webhook, typically the best way to do this is use ngrok_ in order to expose ports externally from your machine.
+  
+See ngrok_ for more information on hosting a webhook from your local machine if you don't have a public address or host.
+
+This is what allows slack to send the messages from it to your bot to get the responses.  Once you put in your webhook address in the OAuth & Permissions section and save it you will have the credentials you need for the slack_credentials.yml file.
+
```
akelad: We don't need this section imo. Further down after the "Directly using python" section I'd change the paragraph that's there to this:

The arguments for the ``HttpInputChannel`` are the port, the url prefix, and the input channel. The default endpoint for receiving messages is ``/webhook``, so the example above would listen for messages on ``/app/webhook``. This is the url you should add in the OAuth & Permissions section. N.b. if you do not set the ``slack_channel`` keyword argument, messages will by delivered back to the user who sent them.

akelad: As for the whole ngrok thing. I think I'd mention this at the very start of this docs page, so change the paragraph there to something like:

Input channels are defined in the ``rasa_core.channels`` module.
 Currently, there is an implementation for the command line as well as
 connection to facebook, slack, telegram, mattermost and twilio.

You need an external webhook to use any of these channels. If you're testing the connection locally, you can set up a temporary webhook using ngrok (link to docs section).

docs/connectors.rst
btotharye: I didn't modify this note, where was it before?

btotharye: ok got this setup back to the way it was

docs/connectors.rst
btotharye: ok have added this as you asked

btotharye:
ok @akelad let me know what you think of these updated changes

akelad:
This looks good now, except I think you've deleted both the Twilio and Mattermost section by accident 😄 as soon as those are back in i can merge

btotharye:
@akelad sorry fixed my screw up :)

akelad:
Thanks a lot!

PR #385, opened 2018-04-16
Dockerfile

### LINKED ISSUE #292: create Dockerfile for rasa core

ricwo:
**Proposed changes**:
- add a Dockerfile that runs the server with `--core` and `--nlu` options
- requires mounting of the model directories

```bash
docker run --mount type=bind,source=<DIALOGUE_SOURCE>,target=/app/model/dialogue --mount type=bind,source=<NLU_MODEL_SOURCE>,target=/app/model/nlu test-core
```

- fixes #292 

- [x] updated the changelog


entrypoint.sh
```diff
@@ -0,0 +1,29 @@
...
+
+case ${1} in
+    start)
+        exec python -m rasa_core.server "${@:2}"
+        ;;
+    run)
+        exec "${@:2}"
+        ;;
+    train)
+        exec python -m rasa_core.train "${@:2}"
```
tmbo: how does the train work because story file & domain are not part of the container nor are they mounted right?

ricwo: Yes, for training you would have to explicitly add `--stories`, `--out` and `--domain`. The files would have to be mounted into the container

Issue #137, opened 2017-12-18
Handling of out of scope messages


buivietan:
Hi there, 
I would like to ask if we can handle the response of the bot if the user input doesn't belong to the list of the intents. For example in restaurant bot if user would ask "I would like to book the flight" the bot can response something like "Sorry I don't understand, can you repeat it again". 
I think we can handle it with the code but is there a easier way to solve this problem? 
Thank you!

tmbo:
There is no default solution for this yet. We are working and thinking about a general applicable solution to the out of scope handling in general.

tmbo:
So to get a fallback for core and nlu here is what I'd do at the moment:
- create a custom nlu interpreter as you said, checking the confidence or doing something else to find out if the message is out of scope and if so, change the intent of the message to e.g. `out_of_scope`
- create a custom `PolicyEnsemble`. e.g. extending `SimplePolicyEnsemble` and overriding `probabilities_using_best_policy` with something along the lines of
```
    def probabilities_using_best_policy(self, tracker, domain):
        result = None
        max_confidence = -1
        
        for p in self.policies:
            probabilities = p.predict_action_probabilities(tracker, domain)
            confidence = np.max(probabilities)
            if confidence > max_confidence:
                max_confidence = confidence
                result = probabilities

        if (tracker.latest_message.intent["name"] == "out_of_scope" 
                or (result and np.max(result) < 0.3)):
            # 0.3 is the probability cutoff for rasa core in this case
            # so if the trained policy returned the most likely action with
            # a probability of less than 0.3, we will run the fallback action
            # instead
            fallback_idx = domain.index_for_action("fallback")
            return utils.one_hot(fallback_idx, domain.num_actions)
        return result
```
- create the above mentioned "fallback_action". should be a custom action with something like this as its code:
```
class ActionFallback(Action):

    def name(self):
        return "fallback"

    def run(self, dispatcher, tracker, domain):
        from rasa_core.events import UserUtteranceReverted

        dispatcher.utter_message("Sorry, didn't get that. Try again.")
        return [UserUtteranceReverted()]

```

ctrado18:
Is above recipe everything to do? And plug the custom action to the domain file?
Do I have to write stories with the fallback Action?

Walter-Ullon:
Is there any documentation thoroughly explain these steps? i.e. creating the custom policy ensemble, etc? I attempted the above but I'm probably missing details. A sample would help immensely. Thanks for all the hard work guys.

[...]

PR #390, opened 2018-04-17
WIP Fallback policy

amn41:
**Proposed changes**:
- add policy for triggering a fallback action when nlu or core model confidence is low
- to be used in an ensemble
- if nlu confidence is below `nlu_threshold`, predict fallbackaction with probability 1
- else predict fallback action with probability `core_threshold`

- fixes https://github.com/RasaHQ/rasa_core/issues/137

- [x] updated the changelog


docs/patterns.rst
```diff
@@ -245,4 +245,21 @@ You can set this value in a custom ``Action``:
...
+
+Fallback / Default Actions
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Sometimes you want to fall back to a default action like saying "Sorry, I didn't understand that".
+To do this, add the ``FallbackPolicy`` to your policy ensemble.
+The default action will be executed if the intent recognition has a confidence below ``nlu_threshold``
+or if none of the dialogue policies predict an action with confidence higher than ``core_threshold``
+
+.. code-block:: python
```
tmbo: needs imports

rasa_core/policies/fallback.py
```diff
@@ -0,0 +1,86 @@
...
+from rasa_core.trackers import DialogueStateTracker
+from rasa_core.training.data import DialogueTrainingData
+
+logger = logging.getLogger(__name__)
+
+
+class FallbackPolicy(Policy):
+    SUPPORTS_ONLINE_TRAINING = False
+
+    def __init__(self,
```
tmbo: needs description of parameters

Harish0596:
@amn41  So now we need to simply pass this Fallback policy to Agent like how we are passing Keras and Memoization?

amn41:
yes http://core.rasa.com/master/patterns.html#fallback-default-actions

NikhilBansal21:
@amn41  i implemented the fallback policy it's working fine and thanks for that but i have a question on running the bot after training it on some instance fallback  is being initialising automatically and the phrase is being coming out 3-4 times .
fallback-issue

amn41:
can you please run your bot with `--debug` to see what's going on? 

maybe it's worth adding a story like

```
* default_intent
  - action_fallback
```

to your data so that Rasa Core knows to predict `action_listen` after `action_fallback` 

(the end of the story will have `action_listen` added to it during training)

imsurinder90:
Actually I am running bot with below code:

```
while True:
    a = input("user: ")
    if a == 'stop':
        break
    responses = agent.handle_message(a)
    for response in responses:
        print("bot: ", response['text'])
```
I didn't get how to run bot in **debug** mode.

I took the latest pull and added fallback policy and a story like:
```
* out_of_context <!-- it is an intent defined in nlu.md -->
   - utter_default <!-- utter_default is 'Sorry i didn't get that' -->
```

When I run the bot with an input like `bla bla bla`, the bot says:
"Sorry i didn't get that" 10 times.

I tried debugging into the code and my understanding is:
`action_listen` is at index 0 in list of actions. So in order to return `action_listen` the index returned must be 0
but the below snippet returns index other than 0
   ```
 def predict_next_action(self, tracker, domain):
        # type: (DialogueStateTracker, Domain) -> (float, int)
        """Predicts the next action the bot should take after seeing x.

        This should be overwritten by more advanced policies to use ML to
        predict the action. Returns the index of the next action"""
        probabilities = self.probabilities_using_best_policy(tracker, domain)
        max_index = np.argmax(probabilities) # **It should return 0 but it doesn't**
                domain.action_for_index(max_index).name(),
                probabilities[max_index]))
        return max_index
```
Please help

**Comment by amn41** [comment]
to run in debug mode, you can pass the `--debug` flag to `rasa_core.run`, or in your python code:

```
import logging

logger = logging.getLogger()
logger.setLevel(logging.DEBUG)
```

**Comment by tmbo** [comment]
I think that needs to go into the docs, you *need* to have a story that tells the bot what to do after the fallback action. otherwise, it might run into a loop.

**Comment by imsurinder90** [comment]
@tmbo I tried adding a story to tell the bot to perform next thing after fallback action but it didn't worked.

```
## Fallback story
* out_of_context
  - utter_default
* thank_you{"thanku_msg": "thanks"}
  - utter_thanks
```

**Output is:**

```
user: bla bla bla
bot:  Sorry, I didn't understand that
bot:  Sorry, I didn't understand that
bot:  Sorry, I didn't understand that
bot:  Sorry, I didn't understand that
bot:  Sorry, I didn't understand that
bot:  Sorry, I didn't understand that
bot:  Sorry, I didn't understand that
bot:  Sorry, I didn't understand that
bot:  Sorry, I didn't understand that
bot:  Sorry, I didn't understand that

user: thanks
bot:  Thank you
```

In below method, the value of action_name is `utter_default` instead of `action_listen`  when user types out of context statement, so the comparison always fails and it returns True

```
    @staticmethod
    def should_predict_another_action(action_name, events):
        is_listen_action = action_name == ACTION_LISTEN_NAME
        return not is_listen_action
```
And below is the code where it goes in loop for 10 times and give `utter_default`  because `should_predict_another_action` always returns True and `self.max_number_of_predictions` is 10 by default.
```
        # action loop. predicts actions until we hit action listen
        while (should_predict_another_action
               and self._should_handle_message(tracker)
               and num_predicted_actions < self.max_number_of_predictions):
            # this actually just calls the policy's method by the same name
            action = self._get_next_action(tracker)

            should_predict_another_action = self._run_action(action,
                                                             tracker,
                                                             dispatcher)
            num_predicted_actions += 1

```

**Comment by tmbo** [comment]
Can you try adding a story that directly starts with the fallback instead of yours, e.g. 
```
## run listen after fallback
  - utter_default
* thank_you{"thanku_msg": "thanks"}
  - utter_thanks
```

**Comment by imsurinder90** [comment]
It works fine If I put the input statement `bla bla bla` into intent `out_of_context`  like:

```
user: bla bla bla
bot:  Sorry, I didn't understand that
user: thanks
bot:  Thank you
```

but if the sentence is not into the intent `out_of_context` it returns `utter_default` 10 times which is an issue.

**Comment by NikhilBansal21** [comment]
@imsurinder90 @tmbo  same thing is happening in my case i.e when fallback is occurring it is repeating multiple times
## Generated Story -7574258607929036394 
* In Dialog Fallback
    - action_fallback

**Comment by NicoHood** [comment]
Does this PR fixes this issue? https://github.com/RasaHQ/rasa_nlu/issues/387

**Comment by NicoHood** [comment]
To me it looks like that with a fallback I cannot train the model via the command line anymore. I am used to call something like:
```
python -m rasa_core.train -d domain.yml -s data/stories.md -o models/partybot/partybot_nlu -u models/partybot/partybot_nlu/ --epochs 200 --online
```

But how can I integrate this new python file? Do I need to start the online training from a python script them? I am not sure how to integrate this fallback properly.

akelad:
Yes it needs to be done in a python file, that command only uses the standard policies.
You need to add the fallback action as presented here (probably also include the MemoizationPolicy in the list of policies): http://rasa.com/docs/core/fallbacks/
And then load your data and train your model like it's done here: https://github.com/RasaHQ/rasa_core/blob/master/rasa_core/train.py#L114

bajiya2024:
i recived same problme. any solutions????

PR #395, opened 2018-04-19
add starter pack to tutorial

akelad:
**Proposed changes**:
- add starter pack to the moodbot tutorial

- [ ] updated the changelog

### DIFF
--- a/docs/tutorial_basics.rst
+++ b/docs/tutorial_basics.rst
@@ -131,11 +131,12 @@ Let's run
 
 .. code-block:: bash
 
-   make train-nlu
+   python -m rasa_nlu.train -c nlu_config.yml --data data/nlu_data.md -o models
+   --fixed_model_name nlu --project current --verbose
 
 to train our NLU model. A new directory ``models/current/nlu`` should have been
 created containing the NLU model. Note that ``current`` stands for project name,
-since this is specified in the train command in the Makefile.
+since this is specified in the train command.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


docs/tutorial_basics.rst
```diff
@@ -184,19 +184,19 @@ To train the dialogue model, run:
 
 .. code-block:: bash
 
-   python -m rasa_core.train -s data/stories.md -d domain.yml -o models/dialogue --epochs 300
```
tmbo: I don't think we should hide these.

docs/tutorial_basics.rst
```diff
@@ -27,32 +27,33 @@ We will create a very simple bot that checks our current mood and tries to
 cheer us up if we are feeling sad. It will query our mood and based on our
 reply will respond with a funny image or a message.
 
-Let's start by creating a project folder:
+We created a `starter pack <https://github.com/RasaHQ/starter-pack>`_ to help
```
tmbo: So now we have that code twice? Once in this repo and once in the starter pack?

akelad: what do you mean? the starter pack is just the bare setup of files, the actual moodbot code is not present in there

docs/tutorial_basics.rst
akelad: yeah fair point, i'll just paste the commands from the make file in here

docs/tutorial_basics.rst
tmbo: ah ok, so you still need to create the data files? sounds good :+1: - was just worried we keep that data in two places.

akelad: no no, all good :D this is the starter pack from the live demo haha

PR #410, opened 2018-04-25
dont double-extract slots

### LINKED ISSUE #407: Form Actions fill more than one slot per entity
**Rasa Core version**:
Current Master with Form Fill Actions
(0.9.0a2)

**Python version**: 
2.7 

OSX

**Issue**:
I noticed a shortcoming of the new form fill actions. For a sentence: “I want to travel to Berlin” it will extract the entity `spacy_GPE: Berlin`. Now with the auto slot filling of the PR it will fill both my `origin` and `destination` slot with Berlin instead of asking the next question in the form. 

Perhaps add a check where a single entity can only be assigned to single slot even if several qualify.
Linking the right slot might be tricky with the generic entity names for duckling and spacy. Perhaps using the first “empty” slot might be first baseline.

**Content of domain file** (if used & relevant):
```yaml
slots:
  destination:
    type: text
  origin:
    type: text
```

amn41:
**Proposed changes**:
- fixes #407 (though not a super general solution) cc @hotzenklotz

- [ ] updated the changelog

### DIFF
--- a/rasa_core/actions/forms.py
+++ b/rasa_core/actions/forms.py
@@ -89,10 +89,20 @@ class FormAction(Action):
 
     def get_other_slots(self, tracker):
         requested_slot = tracker.get_slot("requested_slot")
-        slot_events = [f.extract(tracker) for f in self.required_fields()
-                       if isinstance(f, EntityFormField) and
-                       not f.slot_name == requested_slot]
-        return [e for events in slot_events for e in events]
+
+        requested_entity = None
+        for f in self.required_fields():
+            if f.slot_name == requested_slot:
+                requested_entity = getattr(f, 'entity_name', None)
+
+        slot_events = []
+
+        for f in self.required_fields():
+            if isinstance(f, EntityFormField) and \
+                not f.slot_name == requested_slot and \
+                not f.entity_name == requested_entity:
+                slot_events.extend(f.extract(tracker))
+        return slot_events
 
     def get_requested_slot(self, tracker):
         requested_slot = tracker.get_slot("requested_slot")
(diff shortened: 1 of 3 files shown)


hotzenklotz:
@amn41 Thanks for taking the time to fix this so quickly. I just tried it out and had no luck yet. Both my `origin` and `destination` slots were still set simultaneously  :-(

Judging by your code, this workaround will only work if is `requested_slot` is set initially, right?

amn41:
I was on the fence about this, just added a commit where we are limited to extracting one value for each entity type. lmk if that works

hotzenklotz:
Works like a charm now :-)

rasa_core/actions/forms.py
```diff
@@ -89,10 +89,22 @@ def should_request_slot(self, tracker, slot_name):
...
+
+        requested_entity = None
+        for f in self.required_fields():
+            if f.slot_name == requested_slot:
+                requested_entity = getattr(f, 'entity_name', None)
+
+        slot_events = []
+        extracted_entities = {requested_entity}
+
+        for f in self.required_fields():
```
tmbo: this is more of a hotfix to the issue Tom mentioned, right? The first requested field that needs the entity wins, but it might still not be the right one.

tmbo: I think this is fine as a workaround for now but it really doesn't solve the issue.

amn41: yeah agreed - shall we put it in the inbox to find a better solution for this?

tmbo: already there as entity roles :+1:

amn41:
@tmbo fixed the comment :) good to merge then?

PR #500, opened 2018-05-19
Gh477 fallback policy loop

naoko:
**Proposed changes**:
- Fallback policy could trip cirucuit breaker if subsequent intent is
predicted lower than threshold thus added the condition to prevent
predict twice in a row.

- [ ] updated the changelog

### DIFF
--- a/rasa_core/policies/fallback.py
+++ b/rasa_core/policies/fallback.py
@@ -63,6 +63,17 @@ class FallbackPolicy(Policy):
 
         pass
 
+    def should_fallback(self,
+                        nlu_confidence,  # type float
+                        last_action_name  # type: Text
+                        ):
+        # type: (...) -> bool
+        """It should predict fallback action only if
+        a. predicted NLU confidence is lower than ``nlu_threshold`` &&
+        b. last action is NOT fallback action
+        """
+        return nlu_confidence < self.nlu_threshold and last_action_name != self.fallback_action_name  # noqa
+
     def predict_action_probabilities(self, tracker, domain):
         # type: (DialogueStateTracker, Domain) -> List[float]
         """Predicts a fallback action if NLU confidence is low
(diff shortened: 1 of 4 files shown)


imsurinder90:
FYI, There is an easiest way to fix the issue where fallback policy is repeated number of times. The below code will run only if the `latest_action_name == 'action_listen'` 

```
        score = self.core_threshold
        from rasa_core.actions.action import ACTION_LISTEN_NAME
        if tracker.latest_action_name == ACTION_LISTEN_NAME:
            if nlu_confidence < self.nlu_threshold:
                logger.debug("NLU confidence {} is lower "
                             "than NLU threshold {}. "
                             "Predicting fallback action: {}"
                             "".format(nlu_confidence, self.nlu_threshold,
                                       self.fallback_action_name))
                score = 1.0
        result[idx] = score

        return result
```
Thanks

naoko:
@imsurinder90 , Thank you, I will update accordingly

rasa_core/policies/fallback.py
```diff
@@ -63,6 +63,17 @@ def train(self,
...
+    def should_fallback(self,
+                        nlu_confidence,  # type float
+                        last_action_name  # type: Text
+                        ):
+        # type: (...) -> bool
+        """It should predict fallback action only if
+        a. predicted NLU confidence is lower than ``nlu_threshold`` &&
+        b. last action is NOT fallback action
+        """
+        return nlu_confidence < self.nlu_threshold and last_action_name != self.fallback_action_name  # noqa
```
tmbo: I don't think we should put a `noqa` here.

naoko: Got it. Removed #qa and broke up into 2 lines to comply PEP8

paulhopkins:
I have also had issues with the fallback repeating, so I tried the two suggestions here. However, I found it would cycle between `utter_default` and another action, or would jump about to other parts of the conversation. Can we force it to perform the action and then listen using something like `tracker.trigger_follow_up_action(ActionListen())`. Would this need a custom action?

PR #517, opened 2018-05-23
Add examples missing dependecies

reza-ebrahimi:
Training of `moodbot` example failed to missing dependencies `sklearn-crfsuite` and `spacy`.


tmbo:
Thats right, you need to install `pip install rasa_nlu[spacy]` I made that clear in the documentation of that example, but we can't add spacy to the requirements as it is not a dependency of rasa core per se (you can also only run with keras & tensorflow).

reza-ebrahimi:
Ok, that's reasonable for `spaCy`, what about `sklearn-crfsuite`? 

I have a suggestion that every example project clould have its own `requirements.txt` file to install its dependencies, if `rasa_nalu` already installed it will discards.

Or a mechanism to warn user that `rasa_nlu` was not installed on machine.

tmbo:
I like the idea of having separate requirements in the examples :+1: I'll close this PR, but if you like you can create a requirements in the example.

PR #520, opened 2018-05-23
list slot fix

### LINKED ISSUE #171: Improve (List) Slots
**Problem**

The list slot type only supports single entities as of now. 
This is also described by #165 and #41

1. What happens is that setting the slot overwrites the value in the given slot. Say we have a sentence like "I want to play badminton on Monday and Friday", then we'll have two slot set events fired. One for Monday and one for Friday. The result is that Monday gets written into the slot first and Friday overwrites it. Thus, only Friday remains as a value.

**Proposed Solutions**
1. We could introduce `value` as a property on a slot and then overwrite the setter/getter in subclasses such as `ListSlot` to append new values. This would allow us to solve the above problem, and multiple entities of the same type would get added to the list. 

2. However, I think we ultimately need more advanced slot manipulation methods. For the list slot these could include: set, reset, append, remove. Imagine the following dialogue as a context:
```
User: I would like to play badminton on Monday and Friday
Bot: Ok, I noted down Monday and Friday, should I book courts for you?
```
* Set Action: (sets the value explicitly)
```
User: Actually, that was wrong, I meant I would like to play Tuesday
Bot: Ok, Tuesday then, should I book it?
```
* Reset: (deletes the value)
```
User: Actually, I cannot make it this week
Bot: Ok, too bad. Is there anything else I can do for you?
```
* Append (add a value)
```
User: I could make Wednesday as well!
Bot: Ok, I got Monday, Wednesday, and Friday then. Is that correct?
```
* Remove (remove a value)
``` 
User: Oh wait, I can't make it Friday!
Bot: Alright, so you want to play on Monday then?
```

We could define specific events for these. They could be triggered based on:
1. intents, for example
* play_badminton (set)
* cancel_play (reset)
* add_days(add)
* remove_days(remove)
AND
2. current state of the list slot 

To use intents the user would need to define which intents connect to which slot manipulation action. This would require additional capabilities for bot the domain yaml and the training data format. 

Any thoughts? @tmbo @amn41

akelad:
**Proposed changes**:
- have the list slot actually store a list of entities
- fixes #171 

- [x] updated the changelog

### DIFF
--- a/rasa_core/domain.py
+++ b/rasa_core/domain.py
@@ -295,10 +295,20 @@ class Domain(with_metaclass(abc.ABCMeta, object)):
 
     def slots_for_entities(self, entities):
         if self.store_entities_as_slots:
-            return [SlotSet(entity['entity'], entity['value'])
-                    for entity in entities
-                    for s in self.slots
-                    if entity['entity'] == s.name]
+            slot_events = []
+            listslot_vals = collections.defaultdict(list)
+            for entity in entities:
+                for s in self.slots:
+                    if entity['entity'] == s.name:
+                        if s.type_name == 'list':
+                            listslot_vals[entity['entity']].append(
+                                                              entity['value'])
+                        else:
+                            slot_events.append(SlotSet(entity['entity'],
+                                                       entity['value']))
+            for key, val in listslot_vals.items():
+                slot_events.append(SlotSet(key, val))
+            return slot_events
         else:
             return []
 
--- a/rasa_core/domain.py
+++ b/rasa_core/domain.py
@@ -297,8 +297,9 @@ class Domain(with_metaclass(abc.ABCMeta, object)):
         if self.store_entities_as_slots:
             slot_events = []
             for s in self.slots:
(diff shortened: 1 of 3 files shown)


rasa_core/domain.py
```diff
@@ -295,10 +295,17 @@ def states_for_tracker_history(self, tracker):
 
     def slots_for_entities(self, entities):
         if self.store_entities_as_slots:
-            return [SlotSet(entity['entity'], entity['value'])
-                    for entity in entities
-                    for s in self.slots
-                    if entity['entity'] == s.name]
+            slot_events = []
+            for s in self.slots:
+                matching_entities = [e['value'] for e in entities if
```
tmbo: lets line break like this
```
[e['value'] 
 for e in entities 
 if e['entity'] == s.name]
```

tmbo:
two small things, after that please merge :+1:

anaportela:
Hi! How can I do a "get.slot" for a list? That can "get" one or two different values

PR #522, opened 2018-05-23
Added the reply message in the parse def of server.py

Priprocks:
**Proposed changes**:
- Added the respond in parse def. It saves a http request as I needed both the next action and its response.

- [ ] updated the changelog


Priprocks:
So sorry, I committed with wrong email address.
I am going to send a new pull request

PR #523, opened 2018-05-23
Added the reply message in the parse def of server.py

Priprocks:
**Proposed changes**:
- Added the respond in parse def. It saves a http request as I needed both the next action and its response.

- [ ] updated the changelog


Priprocks:
I do not understand. Why is the build failing?

amn41:
does the [respond](https://core.rasa.com/master/http.html#post--conversations-(str-sender_id)-respond) endpoint do what you want?

Priprocks:
@amn41 Yes, but I modified the next action endpoint to add the output of response.
This way if both the information are needed, it can be done with a single http request.

tmbo:
I don't think it is a good idea to combine these two as the use cases are completly different.

The first API call just predicts the next action, wheras the second one does all predictions until the next action listen and collects all the utterances while doing so.

PR #524, opened 2018-05-23
Read multiple slots from stories

Ghostvv:
**Proposed changes**:
- add ability to read `- slot{"city": "text", "state": "text"}` from stories

- [ ] updated the changelog

### DIFF
--- a/rasa_core/events/__init__.py
+++ b/rasa_core/events/__init__.py
@@ -362,12 +362,22 @@ class SlotSet(Event):
 
     @classmethod
     def _from_story_string(cls, parameters):
-        slot_key = first_key(parameters, default_key="name")
+        slots = []
+        for slot_key, slot_val in parameters.items():
+            if slot_key != "name":
+                slots.append(SlotSet(slot_key, slot_val))
 
-        if slot_key:
-            return SlotSet(slot_key, parameters[slot_key])
+        if slots:
+            return slots
         else:
             return None
+
+        # slot_key = first_key(parameters, default_key="name")
+        #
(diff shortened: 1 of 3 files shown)


rasa_core/events/__init__.py
```diff
@@ -362,10 +362,14 @@ def as_story_string(self):
...
     def _from_story_string(cls, parameters):
-        slot_key = first_key(parameters, default_key="name")
-
-        if slot_key:
-            return SlotSet(slot_key, parameters[slot_key])
+        slots = []
+        for slot_key, slot_val in parameters.items():
+            slots.append(SlotSet(slot_key, slot_val))
+
+        if len(slots) == 1:
```
tmbo: do we need this, can't we always return a list?

rasa_core/training/dsl.py
```diff
@@ -305,11 +305,17 @@ def add_user_messages(self, messages, line_num):
         self.current_step_builder.add_user_messages(parsed_messages)
 
     def add_event(self, event_name, parameters):
-        if "name" not in parameters:
+
+        if "name" not in parameters and event_name != SlotSet.type_name:
```
tmbo: this is a bit hacky, but I don't see any other way right now :+1:

please add a comment ala "need to do this ... slot might be named name ...."

rasa_core/events/__init__.py
Ghostvv: we can, i added this, because i thought it was a problem in tests, but it was actually `name`. I'll change that

PR #525, opened 2018-05-23
Handle request exceptions in model upload

ricwo:
**Proposed changes**:
- wrap `requests.post()` in `try` block

- [ ] updated the changelog

### DIFF
--- a/rasa_core/remote.py
+++ b/rasa_core/remote.py
@@ -11,6 +11,7 @@ import time
 
 import requests
 from future.moves.urllib.parse import quote_plus
+from requests.exceptions import RequestException
 from typing import Callable, Union
 from typing import Text, List, Optional, Dict, Any
 
@@ -144,7 +145,7 @@ class RasaCoreClient(object):
                 if response.status_code == 200:
                     logger.debug("Finished uploading")
                     return response.json()
-            except Exception as e:
+            except RequestException as e:
                 logger.warn("Failed to send model upload request. "
                             "{}".format(e))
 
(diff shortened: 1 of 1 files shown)


rasa_core/remote.py
```diff
@@ -137,12 +137,16 @@ def upload_model(self, model_dir, max_retries=1):
...
-                logger.debug("Finished uploading")
-                return response.json()
+            try:
+                with io.open(model_zip, "rb") as f:
+                    response = requests.post(url, files={"model": f})
+
+                if response.status_code == 200:
+                    logger.debug("Finished uploading")
+                    return response.json()
+            except Exception as e:
```
tmbo: I guess should be fine to just catch `RequestException`, right?

rasa_core/remote.py
```diff
@@ -137,12 +137,16 @@ def upload_model(self, model_dir, max_retries=1):
...
+                logger.warn("Failed to send model upload request. "
+                            "{}".format(e))
 
             if max_retries > 0:
                 # some resting time before we try again - e.g. server
```
tmbo: while you are at it can you put a `if response:` before L156 (infront of the logger)?

rasa_core/remote.py
ricwo: I think it would be `ConnectionError` if anything?

ricwo: although that covers it

tmbo: don't care to much, but would avoid to catch all exceptions.

PR #534, opened 2018-05-25
remove topics

### LINKED ISSUE #496: Remove topics
Topics are not used anymore and so they should be removed from the code base.

akelad:
**Proposed changes**:
- fixes #496 
- remove topics as they're no longer used
- the test tests/test_dialogue_serialisation passes, which has a TopicSet event in it (meaning event parsing won't fail if old TopicSet events are found)

- [x] updated the changelog


rasa_core/events/__init__.py
```diff
@@ -277,6 +277,7 @@ def _from_parameters(cls, parameters):
             raise ValueError("Failed to parse bot uttered event. {}".format(e))
 
 
+# TODO: DEPRECATED - remove in version 0.10.0
 # noinspection PyProtectedMember
 class TopicSet(Event):
```
tmbo: you also need to fix the `apply_to ` method, it shouldn't do anything

akelad: done!

tmbo:
:rocket:

PR #550, opened 2018-05-29
Paper eval scripts

akelad:
**Proposed changes**:
- depends on #494 
- add training/eval scripts from paper 

- [x] updated the changelog


amn41:
@akelad what's the status on this?

m90:
This seems to be already mentioned in the ["Evaluating and Testing" docs](http://www.rasa.com/docs/core/evaluation/#comparing-policies), but is not available with the lastest version on pip / master. Will this land anytime soon or should the section be removed from the docs?

akelad:
oops, those docs should not have been merged already 😄 but yes this PR will be merged soon (ideally today) and then the docs will be relevant

tmbo:
mhm we should rename them though, because paper is not the best name 😅

akelad:
yep i know 😄 i'm still working on making them better atm. do you have any suggestions as to what to name them though because i have no idea

akelad:
I still need to update the docs, but @tmbo the code is ready for a first review

akelad:
@tmbo the idea is to be able to reproduce the results from the paper. and to decide whether to use the embedding policy over the keras policy on your own data i guess. I'm not sure whether we should maybe put these in the generalisation bot repo we're going to open source instead though

tmbo:
I think reproducing results should go in a separate repo. comparing policies to select the best one is something we could add to rasa_core

akelad:
well then this should go in the core repo. because it does compare the two policies on your own dataset by plotting graphs like in the paper

rasa_core/training/dsl.py
```diff
@@ -132,13 +132,22 @@ def __init__(self, domain, interpreter, template_vars=None):
...
+                         exclusion_percentage=None):
         """Given a path reads all contained story files."""
 
         story_steps = []
         for f in nlu_utils.list_files(resource_name):
             steps = StoryFileReader.read_from_file(f, domain, interpreter,
                                                    template_variables)
+            # if exclusion file is present, exclude a percentage of randomly
+            # selected stories
+            if exclusion_file and exclusion_percentage is not 0:
```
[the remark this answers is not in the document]
akelad: @tmbo actually i we want to not have this exclusion file thing, then we need to change the percentages to not start from 0 😄 so should i change it or no?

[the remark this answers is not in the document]
tmbo: I'd say the exclusion file is not really necessary for this use case, the exclusion_percentage should be enough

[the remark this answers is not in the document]
akelad: alright!

rasa_core/evaluate.py
```diff
@@ -81,6 +81,16 @@ def create_argument_parser():
             help="If a prediction error is encountered, an exception "
                  "is thrown. This can be used to validate stories during "
                  "tests, e.g. on travis.")
+    parser.add_argument(
+            '--mode',
+            type=str,
+            default='default',
```
tmbo: this shouldn't be a string but rather a "choice":
`'--mode', choices=('default', 'compare'), help= ...`

rasa_core/evaluate.py
```diff
@@ -383,16 +394,14 @@ def collect_story_predictions(
                          report, precision, f1, accuracy,
                          include_report=False)
 
-    return story_eval_store, failed
+    return story_eval_store, failed, num_stories
```
tmbo: `# type:` annotation needs to be changed

rasa_core/evaluate.py
```diff
@@ -56,21 +63,14 @@ def create_argument_parser():
...
-                 "specified, no plot will be generated.")
+            default="results",
+            help="output path for the any files created from the evaluation")
     parser.add_argument(
             '--e2e', '--end-to-end',
             action='store_true',
             help="Run an end-to-end evaluation for combined action and "
                  "intent prediction. Requires a story file in end-to-end "
                  "format.")
-    parser.add_argument(
```
tmbo: needs to be added to the `migration.rst`

rasa_core/evaluate.py
```diff
@@ -56,21 +63,14 @@ def create_argument_parser():
     parser.add_argument(
             '-o', '--output',
             type=str,
-            nargs="?",
```
tmbo: needs to be added to the migration.rst

rasa_core/evaluate.py
```diff
@@ -485,18 +556,37 @@ def plot_story_evaluation(test_y, predictions,
...
-                         _agent,
-                         cmdline_args.max_stories,
-                         cmdline_args.failed,
-                         cmdline_args.output,
-                         cmdline_args.fail_on_prediction_errors,
-                         cmdline_args.e2e)
+    if cmdline_args.output:
+        nlu_utils.create_dir(cmdline_args.output)
+
+    if cmdline_args.mode == 'default':
```
tmbo: given that we do two different things, I think it makes sense to use subparsers to better group the arguments https://pymotw.com/2/argparse/#nesting-parsers

rasa_core/train.py
```diff
@@ -189,6 +224,54 @@ def train_dialogue_model(domain_file, stories_file, output_path,
...
+                            epochs,
+                            output_path=None,
+                            exclusion_percentage=None,
+                            starspace=True,
+                            max_history=None):
+
+    """Trains either a KerasPolicy model or an EmbeddingPolicy, excluding a
+    certain percentage of a story file"""
+
+    if starspace:
```
tmbo: instead of hardcoding this here, can we maybe use this PR to define different models in different files: https://github.com/RasaHQ/rasa_core/pull/1169

akelad: so you mean pass as an argument different model configs?

tmbo: could be an option: so instead of only passing in one policy.yml (or whatever we are going to call this) specify multiple and compare these

akelad: ok yeah that's a good idea :)

rasa_core/policies/ensemble.py
```diff
@@ -208,40 +202,42 @@ def from_dict(cls, dictionary):
...
-            if policy_name == 'KerasPolicy':
-                policy_object = KerasPolicy(MaxHistoryTrackerFeaturizer(
-                                BinarySingleStateFeaturizer(),
-                                max_history=policy.get('max_history', 3)))
-            else:
-                constr_func = utils.class_from_module_path(policy_name)
-                policy_object = constr_func(**policy)
+            if policy.get('featurizer'):
+                # policy can have only 1 featurizer
+                if len(policy['featurizer']) > 1:
```
tmbo: can we put these into separate functions to reduce the complexity here?

rasa_core/train.py
```diff
@@ -189,82 +177,171 @@ def train_dialogue_model(domain_file, stories_file, output_path,
...
+def train_comparison_models(story_filename,
+                            domain,
+                            output_path=None,
+                            exclusion_percentages=None,
+                            policy_configs=None,
+                            runs=None,
+                            dump_stories=False,
+                            kwargs=None):
+
+    """Trains either a KerasPolicy model or an EmbeddingPolicy, excluding a
```
tmbo: this is outdated. also please stick to the format:
```
def f():
  """Summary line describing the function.

  More explanation in
  a longer text blob."""
```

rasa_core/train.py
```diff
@@ -189,82 +177,171 @@ def train_dialogue_model(domain_file, stories_file, output_path,
...
+    certain percentage of a story file"""
+    for r in range(cmdline_args.runs):
+        logging.info("Starting run {}/{}".format(r + 1, cmdline_args.runs))
```
tmbo: is this relevant enough for the user to be `info` or should it be `debug`

rasa_core/train.py
```diff
@@ -189,82 +177,171 @@ def train_dialogue_model(domain_file, stories_file, output_path,
...
+                            domain, stories, output,
+                            policy_config=policy_config,
+                            exclusion_percentage=i,
+                            kwargs=kwargs,
+                            dump_stories=dump_stories)
+
+
+def get_no_of_stories(stories, domain):
+
+    """gets number of stories in a file"""
```
tmbo:
```suggestion
    """Get number of stories in a file."""
```

rasa_core/train.py
```diff
@@ -189,82 +177,171 @@ def train_dialogue_model(domain_file, stories_file, output_path,
...
+            raise ValueError("You can only pass one config file at a time")
+
+        _agent = train_dialogue_model(domain_file=cmdline_args.domain,
+                                      stories_file=stories,
+                                      output_path=cmdline_args.out,
+                                      dump_stories=cmdline_args.dump_stories,
+                                      policy_config=cmdline_args.config[0],
+                                      kwargs=additional_arguments)
+
+    elif cmdline_args.mode == 'interactive':
```
tmbo: can we put the different modes into separate functions to not put to much code directly into `if __name__ = "main"`

rasa_core/train.py
```diff
@@ -189,82 +177,171 @@ def train_dialogue_model(domain_file, stories_file, output_path,
...
+                                                _endpoints.tracker_store,
+                                                _broker)
+            _agent = Agent.load(cmdline_args.core,
+                                interpreter=_interpreter,
+                                generator=_endpoints.nlg,
+                                tracker_store=_tracker_store,
+                                action_endpoint=_endpoints.action)
+        else:
+            if not cmdline_args.out:
+                raise ValueError("you must provide a path where the model "
```
tmbo:
```suggestion
                raise ValueError("You must provide a path where the model "
```

rasa_core/train.py
```diff
@@ -189,82 +177,171 @@ def train_dialogue_model(domain_file, stories_file, output_path,
...
+
+    elif cmdline_args.mode == 'interactive':
+        _endpoints = AvailableEndpoints.read_endpoints(cmdline_args.endpoints)
+        _interpreter = NaturalLanguageInterpreter.create(cmdline_args.nlu,
+                                                         _endpoints.nlu)
+        if (isinstance(cmdline_args.config, list) and
+                len(cmdline_args.config) > 1):
+            raise ValueError("You can only pass one config file at a time")
+        if cmdline_args.core and cmdline_args.finetune:
+            raise ValueError("--core can only be used without --finetune flag.")
```
tmbo: you need to split the line though
```suggestion
            raise ValueError("The flag `--core` to specify a rasa core "
                             "model can not be used together with "
                             "the `--finetune` flag.")
```

rasa_core/policies/keras_policy.py
```diff
@@ -29,21 +32,32 @@ class KerasPolicy(Policy):
...
                  ):
         # type: (...) -> None
+        if not featurizer:
+            featurizer = self._standard_featurizer(max_history)
         super(KerasPolicy, self).__init__(featurizer)
 
-        self.rnn_size = self.defaults['rnn_size']
-
+        self._load_params(**kwargs)
+        print(self.epochs)
```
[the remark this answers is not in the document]
akelad: 😱

rasa_core/evaluate.py
```diff
@@ -442,20 +460,19 @@ def collect_story_predictions(
                          report, precision, f1, accuracy,
                          in_training_data_fraction,
                          include_report=False)
-
-    return StoryEvalution(evaluation_store=story_eval_store,
-                          failed_stories=failed,
-                          action_list=action_list,
-                          in_training_data_fraction=in_training_data_fraction)
+# TODO: return num_stories somewhere
```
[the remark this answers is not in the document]
akelad: well, i think you reviewed a slightly older version of this PR :P it's done anyways

rasa_core/train.py
akelad: i'd say so, yes. knowing progress of training is standard info, shouldn't need the debug flag for that

PR #557, opened 2018-05-30
fixed error caused by switched arguments in run_story_evaluation

### LINKED ISSUE #545: Wrong function call in evaluation.py
**Rasa Core version**:
0.9.0
**Python version**: 
3.6.4
macOS 10.13.4
**Issue**:
In module `evaluate.py`, line 228, there is a call to `run_story_evaluation`, but I think the parameters are not correctly passed. In particular, the fourth parameter is `max_stories` but `cmdline_args.failed` is passed instead. This leads to errors when trying to launch the evaluation.

RolandJAAI:
**Proposed changes**:
- the arguments cmdline_args.max_stories and cmdline_args.failed are in the wrong order in the call of run_story_evaluation() which leads to "failed_stories.txt" being passed as value for max_stories, therefore the evaluation fails
- this PR fixes it

fixes #545 

- [x] updated the changelog

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,7 @@ Removed
 
 Fixed
 -----
+- fixed evaluation script
 
 [0.9.3] - 2018-05-30
 ^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


RolandJAAI:
@akelad I just noticed this fix is also part of #550  - I just ran into this today as well and fixed it with this PR. Depending on how long you think it may take to merge your PR, maybe you can merge this as a standalone fix earlier, because the evaluation just fails completely right now. Cheers, Roland

tmbo:
Thanks a lot :+1: fixes #545

PR #568, opened 2018-06-04
Track data

JoeyFaulkner:
**Proposed changes**:
- track which story blocks created a training tracker in tracker.block_history dict, including which checkpoints they went through to get there

- [ ] updated the changelog

### DIFF
--- a/rasa_core/featurizers.py
+++ b/rasa_core/featurizers.py
@@ -382,10 +382,8 @@ class TrackerFeaturizer(object):
          trackers_as_actions,
          trackers_hist) = self.training_states_and_actions(trackers,
                                                                  domain)
-
         X, true_lengths = self._featurize_states(trackers_as_states)
         y = self._featurize_labels(trackers_as_actions, domain)
-        trackers_hist = np.array(trackers_hist)
 
         return DialogueTrainingData(X, y, true_lengths, histories=trackers_hist)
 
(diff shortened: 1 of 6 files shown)


rasa_core/featurizers.py
```diff
@@ -598,22 +608,26 @@ def prediction_states(self,
...
+            trackers_histories  #type: List[Dict[int, tuple[text,text]]]
     ):
         # type: (...) -> Tuple[List[List[Dict]], List[List[Dict]]]
         """Removes states that create equal featurizations.
 
         From multiple states that create equal featurizations
         we only need to keep one."""
 
-        hashed_featurizations = set()
+        hashed_featurizations = []
```
JoeyFaulkner: this may be a problem in terms of memory and speed, as sets are significantly faster to lookup than  lists. it could however also be insignificant with respect to other points within the training mechanism

JoeyFaulkner: it seems like this is a definite problem, will have a look at better ways of doing this today

JoeyFaulkner: added a fix which uses a set and a dictionary to perform the list's functionality. slight memory cost and large gain in efficiency

tmbo:
@JoeyFaulkner I don't think we need this at the moment, right?

JoeyFaulkner:
@tmbo nope, was deprioritised!

PR #569, opened 2018-06-04
Added comments in server.py

Priprocks:
- [ ] updated the changelog

### DIFF
--- a/rasa_core/server.py
+++ b/rasa_core/server.py
@@ -24,6 +24,12 @@ from rasa_core.tracker_store import TrackerStore
 from rasa_core.trackers import DialogueStateTracker
 from rasa_core.version import __version__
 
+from typing import Union
+import typing
+
+if typing.TYPE_CHECKING:
+    from rasa_core.interpreter import NaturalLanguageInterpreter as NLI
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


rasa_core/server.py
```diff
@@ -146,25 +146,26 @@ def _create_agent(
...
-               interpreter=None,
-               loglevel="INFO",
-               logfile="rasa_core.log",
-               cors_origins=None,
-               action_factory=None,
-               auth_token=None,
-               tracker_store=None
+def create_app(model_directory,           # type: Text
+               interpreter=None,          # type: Optional
+                                        # [Union(text,NL interpreter)]
```
tmbo: this isn't a proper type hint, should be `Union[Text, NLI, None]` with a

```
if typing.TYPE_CHECKING:
    from rasa_core.interpreter import NaturalLanguageInterpreter as NLI
```
at the top (after the imports).

additionally you need to add 
```
from typing import Union
import typing
```
to the imports

tmbo: please also don't align the type hints but rather put them two spaces after the argument.

rasa_core/server.py
```diff
@@ -146,25 +146,26 @@ def _create_agent(
...
     app = Flask(__name__)
     CORS(app, resources={r"/*": {"origins": "*"}})
-
+    # Setting up logfile
     utils.configure_file_logging(loglevel, logfile)
 
     if not cors_origins:
         cors_origins = []
-
+    type(cors_origins)
```
tmbo: I guess this is not needed

Priprocks: Okay sure! I'm making the necessary changes as you said

Priprocks:
Did the changes as you asked. Thank you for helping me out! :)

rasa_core/server.py
```diff
@@ -146,25 +152,24 @@ def _create_agent(
...
-def create_app(model_directory,
-               interpreter=None,
-               loglevel="INFO",
-               logfile="rasa_core.log",
-               cors_origins=None,
-               action_factory=None,
-               auth_token=None,
-               tracker_store=None
+def create_app(model_directory,  # type: Text
+               interpreter=None,  # type: Optional[Union(text,NL interpreter)]
```
tmbo: should be `# type: Union[Text, NLI, None]`

tmbo:
Thanks :rocket:

PR #587, opened 2018-06-07
Caching for featurizer

tmbo:
**Proposed changes**:
- ...

- [ ] updated the changelog

### DIFF
--- a/rasa_core/trackers.py
+++ b/rasa_core/trackers.py
@@ -23,6 +23,7 @@ logger = logging.getLogger(__name__)
 
 if typing.TYPE_CHECKING:
     from rasa_core.actions import Action
+    from rasa_core.domain import Domain
 
 
 class DialogueStateTracker(object):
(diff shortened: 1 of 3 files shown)


rasa_core/trackers.py
```diff
@@ -113,6 +114,13 @@ def current_state(self,
             "events": evts
         }
 
+    def past_states(self, domain):
+        # type: (Domain) -> deque
+        """Generate the past states of this tracker based on the history."""
+
+        generated_states = domain.states_for_tracker_history(self)
+        return deque((frozenset(s.items()) for s in generated_states))
```
Ghostvv: I'm confused a bit now, will it cache the probabilities as well?

tmbo: yes it will cache tuples of the name and the value, so `frozenset([(u'intent_greet', 1.0), (u'prev_action_listen', 1.0)])`

Ghostvv: Thanks, just wanted to make sure, I get it correctly

PR #592, opened 2018-06-08
Added support for dialogflow emulator

Priprocks:
**Proposed changes**:
- ...

- [ ] updated the changelog


tmbo:
Thanks a lot for this suggestion. After thinking about this for quite some time, I do not think we want to go down this road at this time. Dialogflow has a very different concept about who conversations work and this concept is not to compatible with the way rasa core works. 

Creating endpoints to support dialogflow based code would suggest though, that you can simply exchange a dialogflow bot with rasa core - but that transition involves more effort, because you are switching from a rule based model to an ML based on. Therefore, I'd like to make that explicit by not supporting a dialogflow based API.

PR #593, opened 2018-06-08
Event stream handler

### LINKED ISSUE #554: Create event stream output
Rasa Core should submit new events on trackers to an output (e.g. Kafka) as an event stream. This allows other systems to consume the data about conversations and do some analytics.

ricwo:
**Proposed changes**:
- fixes #554 
- creates a RabbitMQ-based event stream handler using Pika

- [x] updated the changelog


docs/brokers.rst
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+The ``event`` field takes the event's ``type_name`` (for more on event
+types, check out the :doc:`api/events` docs). You need to have a RabbitMQ
+server running, as well as another application
+that consumes the events. This consumer to needs to implement Pika's
+``start_consuming()`` method with a ``callback`` action. Here's a simple
+example:
+
+.. code-block:: python
```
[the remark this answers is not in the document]
tmbo: I don't think we need the class though. for the example its fine to put the connection stuff into main and only have a function for the callback. 

I don't think the queue is called `my_events` though right?

rasa_core/broker.py
```diff
@@ -0,0 +1,43 @@
...
+
+import logging
+
+import pika
+
+logger = logging.getLogger(__name__)
+
+
+class EventBroker(object):
+    def process(self, body):
```
tmbo: Should get a short description (and I'd name the parameter `event` as you did in the producer implementation).

rasa_core/broker.py
```diff
@@ -0,0 +1,43 @@
...
+from __future__ import unicode_literals
+
+import logging
+
+import pika
+
+logger = logging.getLogger(__name__)
+
+
+class EventBroker(object):
```
tmbo: This should rather be called `EventChannel` (it is not a broker from a technical point of view) - I think the correct way to call it would be an event channel connection but that is a horrible name. the broker is kafka.

Accordingly, I'd call `process` smth like `emit` or `publish`

rasa_core/tracker_store.py
```diff
@@ -49,6 +52,18 @@ def retrieve(self, sender_id):
...
         raise NotImplementedError()
 
+    def stream_events(self, tracker):
+        # type: (DialogueStateTracker) -> None
+        old_tracker = self.retrieve(tracker.sender_id)
+        offset = len(old_tracker.events) if old_tracker else 0
+        for evt in tracker.events[offset:]:
+            body = {
+                "sender_id": tracker.sender_id,
+                "name": evt.type_name,
```
tmbo: we are missing the complete event here. I don't think we need `name` but rather the whole `evt`.

tmbo:
please let me know when this is ready for another review. (also please merge master into your branch)

rasa_core/tracker_store.py
ricwo: ready for another review 👍

docs/brokers.rst
```diff
@@ -0,0 +1,78 @@
...
+        connection = pika.BlockingConnection(
+            pika.ConnectionParameters('rabbit', credentials=credentials))
+
+        # start consumption of channel
+        channel = connection.channel()
+        channel.basic_consume(_callback,
+                              queue='rasa_core_events',
+                              no_ack=True)
+        channel.start_consuming()
+
```
tmbo: I guess we don't need all these newlines :D

tmbo:
should be merged into `sdk` though instead of master.

PR #600, opened 2018-06-08
Fix of augmented memoization policy

Ghostvv:
**Proposed changes**:
- make augmented memoization policy recall memorized training stories if they were defined as separate stories, but not more

- [x] updated the changelog


Ghostvv:
Should be merged after https://github.com/RasaHQ/rasa_core/pull/591

PR #631, opened 2018-06-19
wrap scheduler init in try/except

amn41:
**Proposed changes**:
- wrap initialisation of backgroundscheduler in a try/ excepy
- allows rasa core to run in docker containers without first setting the system timezone (e.g. binderhub, jupyterhub, colaboratory)

- [x] updated the changelog

### DIFF
--- a/rasa_core/processor.py
+++ b/rasa_core/processor.py
@@ -9,6 +9,7 @@ import warnings
 from types import LambdaType
 
 from apscheduler.schedulers.background import BackgroundScheduler
+from pytz import UnknownTimeZoneError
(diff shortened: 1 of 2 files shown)


rasa_core/processor.py
```diff
@@ -29,11 +29,17 @@
...
-scheduler = BackgroundScheduler()
-scheduler.start()
+
 
 logger = logging.getLogger(__name__)
 
+try:
+    scheduler = BackgroundScheduler()
+    scheduler.start()
+except:
```
tmbo: can we catch a more specific exception here? or is it just throwing a generic `Exception`?

PR #643, opened 2018-06-20
Train from url

JoeyFaulkner:
**Proposed changes**:
- add command line argument to be able to fetch trackers from a URL

- [ ] updated the changelog


rasa_core/train.py
```diff
@@ -113,6 +122,12 @@ def train_dialogue_model(domain_file, stories_file, output_path,
                                                  "debug_plots"})
     training_data = agent.load_data(stories_file, **data_load_args)
 
+    if url is not None:
```
tmbo: should go into a separate function (probably this block as well as the line before with the data loading, so that all the data loading is in one function)

rasa_core/train.py
```diff
@@ -113,6 +122,12 @@ def train_dialogue_model(domain_file, stories_file, output_path,
...
+        tracker_store = TrackerStore(TemplateDomain(domain_file))
+        serialised_tracs = requests.get(url).json()
+        url_trackers = [tracker_store.deserialise_tracker(42, trac) for trac in serialised_tracs]
```
tmbo: shouldn't need a tracker store, how about
```
DialogueStateTracker.from_dict(UserMessage.DEFAULT_SENDER_ID,
                                                      serialised_trackers,
                                                      agent.domain.slots)
```

rasa_core/train.py
```diff
@@ -87,6 +90,11 @@ def create_argument_parser():
             action='store_true',
             help="If enabled, save flattened stories to a file")
 
+    parser.add_argument(
+            '--url',
+            default=None,
+            help="If supplied, downloads serialised trackers from a URL and trains them")
```
tmbo: what is the format for the trackers, is it a list of events?

amn41:
@JoeyFaulkner we'll need to update this with the new stories endpoints, http://rasa.com/docs/platform/master/api/data/#operation/getStories

rasa_core/train.py
```diff
@@ -111,6 +121,12 @@ def train_dialogue_model(domain_file, stories_file, output_path,
                                                  "augmentation_factor",
                                                  "remove_duplicates",
                                                  "debug_plots"})
+
+    if url is not None:
+        data_store = NamedTemporaryFile()
```
tmbo: please move into a separate function and add error handling (e.g. `response.status_code != 200`)

PR #645, opened 2018-06-21
Model server queries

ricwo:
**Proposed changes**:
- Agent regularly queries model server

- [ ] updated the changelog

### DIFF
--- a/rasa_core/agent.py
+++ b/rasa_core/agent.py
@@ -6,8 +6,12 @@ from __future__ import unicode_literals
 import logging
 import os
 import shutil
+import time
+from threading import Thread
 
+import requests
 import typing
+import zipfile
 from six import string_types
 from typing import Text, List, Optional, Callable, Any, Dict, Union
 
@@ -22,6 +26,7 @@ from rasa_core.policies.memoization import MemoizationPolicy
 from rasa_core.processor import MessageProcessor
 from rasa_core.tracker_store import InMemoryTrackerStore, TrackerStore
 from rasa_core.trackers import DialogueStateTracker
+from rasa_nlu.utils import is_url, create_temporary_file
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


rasa_core/agent.py
```diff
@@ -41,23 +52,36 @@ def __init__(
             domain,  # type: Union[Text, Domain]
             policies=None,  # type: Union[PolicyEnsemble, List[Policy], None]
             interpreter=None,  # type: Union[NLI, Text, None]
-            tracker_store=None  # type: Optional[TrackerStore]
+            tracker_store=None,  # type: Optional[TrackerStore]
+            model_server=None,  # type: Optional[Text]
```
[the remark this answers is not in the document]
tmbo: that one is merged now, you can import it from `rasa_core.utils`

rasa_core/agent.py
```diff
@@ -419,6 +468,85 @@ def _create_ensemble(policies):
...
             raise ValueError(
-                    "Invalid param `policies`. Passed object is "
-                    "of type '{}', but should be policy, an array of "
-                    "policies, or a policy ensemble".format(passed_type))
+                "Invalid param `policies`. Passed object is "
+                "of type '{}', but should be policy, an array of "
+                "policies, or a policy ensemble".format(passed_type))
+
+    @staticmethod
+    def _get_model_hash(model_server, header='model_hash'):
```
tmbo: can we use a standard header field for this, e.g. the `ETag` header?

tmbo: thinking about this more, we could even just use the `GET` request to retrieve the model, sending a header `If-None-Match: "PREVIOUS_MODEL_HASH"`, the standard says that the server should return `204 - Not modified` with an empty body if the resource has not changed and otherwise should return the resource.

This would move this checking logic to the backend and rasa core would also work with a model storage backend like s3

rasa_core/agent.py
```diff
@@ -419,6 +468,85 @@ def _create_ensemble(policies):
...
+                                  model_directory  # type: Text
+                                  ):
+        # type: (...) -> Optional[(Domain, PolicyEnsemble, Text)]
+        """Loads a zipped Rasa Core model from a URL."""
+
+        if not is_url(model_server):
+            raise requests.exceptions.InvalidURL(model_server)
+        new_hash = Agent._get_model_hash(model_server)
+        if new_hash != self.model_hash:
+            self.model_hash = new_hash
```
tmbo: isn't this duplicating the code from `init_model_from_server`

rasa_core/remote.py
```diff
@@ -280,14 +276,18 @@ def load(cls,
...
         # type: (...) -> RemoteAgent
 
         domain = TemplateDomain.load(os.path.join(path, "domain.yml"),
                                      action_factory)
 
         core_client = RasaCoreClient(core_host, auth_token)
-        core_client.upload_model(path, max_retries=5)
+
+        if upload_model:
+            core_client.upload_model(path, max_retries=max_retries)
```
tmbo: is there a case where you wouldn't want to upload the model?

tests/test_agent.py
```diff
@@ -81,3 +82,23 @@ def test_agent_wrong_use_of_load(tmpdir, default_domain):
...
+
+
+@responses.activate
+def test_agent_with_model_server(tmpdir, zipped_moodbot_model):
+    model_server_url = 'http://server.com/model/default_core@latest'
+    model_hash = 'somehash'
+    responses.add(responses.HEAD,
+                  model_server_url,
+                  headers={"model_hash": "somehash"})
+    with open(zipped_moodbot_model, 'rb') as f:
```
tmbo: io.open

docs/http.rst
```diff
@@ -61,6 +61,24 @@ The different parameters are:
 - ``-u``, which is the path to the Rasa NLU model.
 - ``-o``, which is the path to the log file.
 
+Fetching models from a server
+-----------------------------
+You can also configure the http server to fetch models from another URL:
+
+.. code-block:: bash
+
+    $ python -m rasa_core.server -d examples/babi/models/policy/current -u examples/babi/models/nlu/current_py2 --model_server_url http://my-server.com/models/default_core@latest -o out.log
```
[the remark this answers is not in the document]
ricwo: we still need a directory for saving the model

rasa_core/agent.py
ricwo: apart from the first two lines the methods are quite different. `update_model` assumes state, `init_model` doesn't

PR #676, opened 2018-06-28
fix fallback policy

akelad:
**Proposed changes**:
- give the fallback policy a confidence of 1.1 so it overrides the memo and any other policy (previously the fallback action would not get executed because other policies also had a confidence of 1.0)

- [ ] updated the changelog


rasa_core/policies/ensemble.py
```diff
@@ -218,7 +218,8 @@ def probabilities_using_best_policy(self, tracker, domain):
                 max_confidence = confidence
                 result = probabilities
                 best_policy_name = 'policy_{}_{}'.format(i, type(p).__name__)
-
+        # normalize probablilities
+        result = result / np.linalg.norm(result)
```
tmbo: please check that sum(result) is not 0 first

PR #679, opened 2018-06-29
Domain API endpoint

JoeyFaulkner:
**Proposed changes**:
- create domain from yaml string
- create yaml string from domain
- expose endpoint for fetching domain yaml

- [ ] updated the changelog

### DIFF
--- a/rasa_core/domain.py
+++ b/rasa_core/domain.py
@@ -404,7 +404,7 @@ class TemplateDomain(Domain):
             action_factory = data.get("action_factory", None)
         slots = cls.collect_slots(data.get("slots", {}))
         additional_arguments = data.get("config", {})
-        return TemplateDomain(
+        return cls(
                 data.get("intents", []),
                 data.get("entities", []),
                 slots,
--- a/rasa_core/domain.py
+++ b/rasa_core/domain.py
@@ -496,7 +496,7 @@ class TemplateDomain(Domain):
     def _slot_definitions(self):
         return {slot.name: slot.persistence_info() for slot in self.slots}
 
-    def persist(self, filename):
+    def _make_domain_data(self):
         additional_config = {
             "store_entities_as_slots": self.store_entities_as_slots}
         action_names = self.action_names[len(Domain.DEFAULT_ACTIONS):]
@@ -511,8 +511,15 @@ class TemplateDomain(Domain):
             "action_names": action_names,  # names in stories
             "action_factory": self._factory_name
         }
+        return domain_data
 
+    def persist(self, filename):
+        domain_data = self._make_domain_data()
         utils.dump_obj_as_yaml_to_file(filename, domain_data)
+
+    def to_yaml(self):
+        domain_data = self._make_domain_data()
+        return utils.dump_obj_as_yaml_to_string(domain_data)
 
     @utils.lazyproperty
     def templates(self):
--- a/rasa_core/domain.py
+++ b/rasa_core/domain.py
@@ -420,7 +420,7 @@ class TemplateDomain(Domain):
         )
 
     @classmethod
-    def validate_domain_yaml(cls, input):
+    def validate_domain_yaml(cls, yaml):
         """Validate domain yaml."""
         from pykwalify.core import Core
 
@@ -429,7 +429,7 @@ class TemplateDomain(Domain):
 
         schema_file = pkg_resources.resource_filename(__name__,
                                                       "schemas/domain.yml")
-        source_data = utils.read_yaml_string(input)
+        source_data = utils.read_yaml_string(yaml)
         c = Core(source_data=source_data,
                  schema_files=[schema_file])
         try:
--- a/rasa_core/domain.py
+++ b/rasa_core/domain.py
@@ -28,7 +28,7 @@ from rasa_core.actions.factories import (
     ensure_action_name_uniqueness)
 from rasa_core.slots import Slot
 from rasa_core.trackers import DialogueStateTracker, SlotSet
-from rasa_core.utils import read_yaml_file
+from rasa_core.utils import read_yaml_file, read_yaml_string
 
 logger = logging.getLogger(__name__)
 
@@ -416,7 +416,27 @@ class TemplateDomain(Domain):
         )
 
     @classmethod
-    def validate_domain_yaml(cls, filename):
+    def load_from_yaml(cls, yaml, action_factory=None):
+        cls.validate_domain_yaml(yaml, string_input=True)
+        data = read_yaml_string(yaml)
+        utter_templates = cls.collect_templates(data.get("templates", {}))
+        if not action_factory:
+            action_factory = data.get("action_factory", None)
+        slots = cls.collect_slots(data.get("slots", {}))
+        additional_arguments = data.get("config", {})
+        return TemplateDomain(
+            data.get("intents", []),
+            data.get("entities", []),
+            slots,
+            utter_templates,
+            data.get("actions", []),
+            data.get("action_names", []),
+            action_factory,
+            **additional_arguments
+        )
+
+    @classmethod
+    def validate_domain_yaml(cls, input, string_input=False):
         """Validate domain yaml."""
         from pykwalify.core import Core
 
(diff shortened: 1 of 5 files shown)


tests/test_domain.py
```diff
@@ -166,3 +166,20 @@ def test_domain_fails_on_unknown_custom_slot_type(tmpdir):
...
+config:
+  store_entities_as_slots: true
+entities: []
+intents: []
+slots: {}
+templates:
+  utter_greet:
+  - text: hey there!"""
+    domain = TemplateDomain.load_from_yaml(test_yaml)
+    assert test_yaml in domain.to_yaml()
```
amn41: why 'in' ? would `x.strip() == y.strip()` also work?

rasa_core/utils.py
```diff
@@ -17,6 +17,10 @@
 from builtins import input, range, str
 from numpy import all, array
 from typing import Text, Any, List, Optional, Tuple, Dict, Set
+try:
```
amn41: is this a py2 / py3 thing? can we use `six` instead?

rasa_core/server.py
```diff
@@ -305,6 +305,16 @@ def update_tracker(sender_id):
         agent().tracker_store.save(tracker)
         return jsonify(tracker.current_state(should_include_events=True))
 
+    @app.route("/domain",
+               methods=['GET'])
+    @cross_origin(origins=cors_origins)
+    @requires_auth(auth_token)
+    @ensure_loaded_agent(agent)
+    def get_domain_yaml():
+        """Use a list of events to set a conversations tracker to a state."""
```
amn41: description needs updated

rasa_core/domain.py
```diff
@@ -416,7 +416,27 @@ def load(cls, filename, action_factory=None):
...
-    def validate_domain_yaml(cls, filename):
+    def load_from_yaml(cls, yaml, action_factory=None):
+        cls.validate_domain_yaml(yaml, string_input=True)
+        data = read_yaml_string(yaml)
+        utter_templates = cls.collect_templates(data.get("templates", {}))
+        if not action_factory:
+            action_factory = data.get("action_factory", None)
+        slots = cls.collect_slots(data.get("slots", {}))
+        additional_arguments = data.get("config", {})
+        return TemplateDomain(
```
amn41: why do we have both `cls` and `TemplateDomain` in this method? surely they are the same? if we are allowing for subclassing then probably stick with `cls` ?

rasa_core/utils.py
```diff
@@ -319,6 +323,21 @@ def read_yaml_file(filename):
 
         return yaml_parser.load(read_file(filename))
 
+def read_yaml_string(string):
```
amn41: can we then rewrite `read_yaml_file` to just be `read_yaml_string(read_file(filename))` ?

rasa_core/utils.py
```diff
@@ -319,6 +323,21 @@ def read_yaml_file(filename):
...
+
+    if six.PY2:
```
amn41: here an example of using `six`

rasa_core/utils.py
JoeyFaulkner: it is, use six how?

rasa_core/domain.py
JoeyFaulkner: oh yes, good call

rasa_core/utils.py
JoeyFaulkner: saw your other comment, ignore me!

tests/test_domain.py
JoeyFaulkner: yep, that works fine!

rasa_core/domain.py
```diff
@@ -416,7 +416,27 @@ def load(cls, filename, action_factory=None):
         )
 
     @classmethod
-    def validate_domain_yaml(cls, filename):
+    def load_from_yaml(cls, yaml, action_factory=None):
```
tmbo: can `load` use `load_from_yaml` once it read the yaml from the file as well? want to avoid the duplication

rasa_core/utils.py
```diff
@@ -341,6 +349,27 @@ def dump_obj_as_yaml_to_file(filename, obj):
         with io.open(filename, 'w', encoding="utf-8") as yaml_file:
             yaml_writer.dump(obj, yaml_file)
 
+def dump_obj_as_yaml_to_string(obj):
+    """Writes data (python dict) to a yaml string."""
+
+    if six.PY2:
```
tmbo: these two functions should share the code, eg:
```
def _dump_yaml(obj, output):
    #...
    yaml_writer.dump(obj, output)
    #...

def dump_obj_as_yaml_to_file(filename, obj):
    with io.open(filename, 'w', encoding="utf-8") as yaml_file:
        _dump_yaml(obj, yaml_file)

def dump_obj_as_yaml(obj):
    str_io = StringIO()
    _dump_yaml(obj, str_io)
   return str_io.getvalue()
```

rasa_core/server.py
```diff
@@ -305,6 +305,16 @@ def update_tracker(sender_id):
...
         return jsonify(tracker.current_state(should_include_events=True))
 
+    @app.route("/domain",
+               methods=['GET'])
+    @cross_origin(origins=cors_origins)
+    @requires_auth(auth_token)
+    @ensure_loaded_agent(agent)
+    def get_domain_yaml():
+        """Get current domain in yaml format."""
+        domain_yaml = agent().domain.to_yaml()
```
tmbo: why are we sending the domain as yaml instead of json?

JoeyFaulkner: I mean its the way that the domain is stored in the model folder so why not?

rasa_core/domain.py
```diff
@@ -390,33 +390,37 @@ def templates(self):
...
+            utter_templates,
+            data.get("actions", []),
+            data.get("action_names", []),
+            action_factory,
+            **additional_arguments
         )
 
     @classmethod
-    def validate_domain_yaml(cls, filename):
+    def validate_domain_yaml(cls, input, string_input=False):
```
tmbo: `input` needs to be renamed (shadows build in function)

do we need `string_input`? I think it is always `True`

rasa_core/domain.py
```diff
@@ -496,7 +504,7 @@ def instantiate_actions(factory_name, action_classes, action_names,
     def _slot_definitions(self):
         return {slot.name: slot.persistence_info() for slot in self.slots}
 
-    def persist(self, filename):
+    def _make_domain_data(self):
```
tmbo: `def  as_dict(self)` (for consistency across classes)

rasa_core/domain.py
```diff
@@ -511,9 +519,16 @@ def persist(self, filename):
             "action_names": action_names,  # names in stories
             "action_factory": self._factory_name
         }
+        return domain_data
 
+    def persist(self, filename):
+        domain_data = self._make_domain_data()
         utils.dump_obj_as_yaml_to_file(filename, domain_data)
 
+    def to_yaml(self):
```
tmbo: `as_yaml` for consistency

rasa_core/server.py
tmbo: this is a REST api, so usually the responses are in json format. I mean I don't have a strong opinion on this, but if yaml is used, content negotiation should be used: e.g.

```python
accepts = request.headers.get("Accept", default="application/json")
if accepts.endswith("json"):
  domain = agent().domain.as_dict()
  return jsonify(domain)
elif accepts.endswith("yml"):
  domain = agent().domain.as_yaml()
  return Response(domain_yaml, status=200, content_type="application/x-yml")
else:
  return Response("""Invalid accept header. Domain can be provided as json ("Accept: application/json") or yml ("Accept: application/x-yml"). Make sure you've set the appropriate Accept header.""", status=406)
```

rasa_core/domain.py
JoeyFaulkner: it wasnt in tests but i changed tests :)

rasa_core/domain.py
```diff
@@ -390,33 +390,37 @@ def templates(self):
...
+            utter_templates,
+            data.get("actions", []),
+            data.get("action_names", []),
+            action_factory,
+            **additional_arguments
         )
 
     @classmethod
-    def validate_domain_yaml(cls, filename):
+    def validate_domain_yaml(cls, input):
```
tmbo: still needs to be renamed to something that is not `input` (shadows built in)

PR #707, opened 2018-07-09
Implement send_image_url for CollectingOutputChannel

m90:
**Proposed changes**:
When serializing the return value of Agent.handle_message, image attachments
will currently be serialized into plain text, which makes it hard for 3rd
parties to consume the data. Instead, also add the image_url to `data`.

- [ ] updated the changelog

I'm wrapping `rasa_core` in a Flask app that is consumed via JSON API. When I `jsonify` the bot's response (handled via `CollectingOutputChannel`), images are delivered in a way that makes it very hacky for consumers of my app to identify images.

For example sending buttons will right now look like:

```json
{
    "messages": [
        {
            "data": [
                {
                    "payload": "Yes",
                    "title": "Yes"
                },
                {
                    "payload": "No",
                    "title": "No"
                }
            ],
            "recipient_id": "some-id",
            "text": "Buttons, there will be buttons:"
        }
    ]
}
```

when sending images looks like:

```json
{
    "messages": [
        {
            "recipient_id": "some-id",
            "text": "Image, there will be an image:"
        },
        {
            "recipient_id": "some-id",
            "text": "Image: https://upload.wikimedia.org/wikipedia/commons/4/4b/Schnabeltier2.jpg"
        }
    ]
}
```

This PR adds the attachment data to `data`, which would let consumers check for `data` and then discard `text` if they want to.

I'm not sure if this is the most elegant way to do it, as for example sending buttons will still create a payload consisting of a single message when sending an image creates two, but it's the only non-breaking way I could come up with.

Maybe it'd be a possibility for `0.10` or later to change the signature of `Dispatcher#utter_attachment` from `self, attachment` to `self, text, attachment, **kwargs` to be inline with how buttons behave?


tmbo:
thanks for the suggestion - we changed a lot with respect to the input channels and took your suggestion as input. the collecting channel should now properly handle images / attachements and buttons

m90:
Sounds good, will this be in `0.11` or what is  the target version?

tmbo:
yes merged into master and will be part of 0.11

PR #727, opened 2018-07-11
Allow unicode characters in extracted entity value

tdzienniak:
**Proposed changes**:
- change string parsing function from `str` to `unicode` to allow usage of unicode characters in extracted entity value during interactive training

- [ ] updated the changelog


tmbo:
there is no `unicode` in python 3. can you try using `str` again, but importing `from builtins import str`, does that work for you?

tdzienniak:
Yes, it works when imported from `builtins`.

tmbo:
Great, do you mind updating the PR accordingly?

PR #728, opened 2018-07-11
Allow an external endpoint to produce bot messages

tmbo:
**Proposed changes**:
- moves from the template based generation of bot responses to a more general approach

- [x] updated the changelog

### DIFF
--- a/rasa_core/nlg/callback.py
+++ b/rasa_core/nlg/callback.py
@@ -14,7 +14,7 @@ from rasa_core.utils import EndpointConfig
 logger = logging.getLogger(__name__)
 
 
-def nlg_response_format():
+def nlg_response_format_spec():
     """Expected response schema for an NLG endpoint.
 
     Used for validation of the response returned from the NLG endpoint."""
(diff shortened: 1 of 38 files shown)


imsurinder90:
@tmbo it would be easier to understand the changes if you add comments in newly created file and explain the changes in detail :)

imsurinder90:
@tmbo thanks for adding comments in the new files and also migration guide which makes it quite easy to understand the code.

I want to understand in which use-case I can add endpoint and add to the interpreter. If there is a feature request created for this, can you please provide link to that so I can understand when such implementation is required.

rasa_core/run.py
```diff
@@ -137,15 +144,51 @@ def create_input_channel(channel, port, credentials_file):
             raise Exception("Unknown input channel for running main.")
 
 
+def interpreter_from_args(nlu_model, nlu_endpoint):
```
ricwo: A short description and typing would be helpful

rasa_core/nlg/callback.py
```diff
@@ -0,0 +1,131 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import logging
+
+from typing import Text, Any, Dict
+
+from rasa_core.nlg.generator import NaturalLanguageGenerator
```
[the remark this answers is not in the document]
tmbo: this is inside the package so it is better to use the fullpath here, but I changed it for the other ones :+1:

rasa_core/nlg/callback.py
```diff
@@ -0,0 +1,131 @@
...
+        "type": "object",
+        "properties": {
+            "text": {"type": "string"},
+            "buttons": {
+                "type": "array",
+                "items": {"type": "object"}
+            },
+            "elements": {
+                "type": "array",
+                "items": {"type": "object",}
```
ricwo: @tmbo just noticed this typo

tmbo: :+1:

rasa_core/remote.py
```diff
@@ -80,43 +81,42 @@ def tracker_json(self,
                      ):
         """Retrieve a tracker's json representation from remote instance."""
 
-        url = ("{}/conversations/{}/tracker?token={}"
-               "&ignore_restarts={}"
-               "&events={}").format(self.host, sender_id, self.token,
-                                    use_history, include_events)
+        url = "/conversations/{}/tracker?&ignore_restarts={}&events={}".format(
```
ricwo: the token got lost here

PR #731, opened 2018-07-11
New fallback action

akelad:
**Proposed changes**:
- Create a new default action ActionFallback

- [x] updated the changelog

### DIFF
--- a/docs/fallbacks.rst
+++ b/docs/fallbacks.rst
@@ -16,7 +16,7 @@ or if none of the dialogue policies predict an action with confidence higher tha
    from rasa_core.policies.keras_policy import KerasPolicy
    from rasa_core.agent import Agent
 
-   fallback = FallbackPolicy(fallback_action_name="action_fallback",
+   fallback = FallbackPolicy(fallback_action_name="action_default_fallback",
                              core_threshold=0.3,
                              nlu_threshold=0.3)
 
@@ -27,10 +27,17 @@ or if none of the dialogue policies predict an action with confidence higher tha
 ``action_fallback`` is a default action in Rasa Core, which will send the
 ``utter_default`` template message to the user. Make sure to specify this template
 in your domain file. It will also revert back to the state of the conversation
-before this action was executed, so that it will not influence the prediction of
-future actions. You can take a look at the source of the action below:
+before the user message that caused the fallback, so that it will not influence
+the prediction of future actions. You can take a look at the source of the
+action below:
 
-.. autoclass:: rasa_core.actions.action.ActionFallback
+.. autoclass:: rasa_core.actions.action.ActionDefaultFallback
+
+.. note::
+  You can also create your own custom action to use as a fallback. Be aware
+  that if this action does not return a ``UserUtteranceReverted`` event, the
+  next predictions of your bot may become inaccurate, as it very likely that the
+  fallback action is not present in your stories
 
 If you have a specific intent that will trigger this, let's say it's called ``out_of_scope``, then you
 should add this as a story:
@@ -39,4 +46,4 @@ should add this as a story:
 
     ## fallback story
     * out_of_scope
-      - action_fallback
+      - action_default_fallback
--- a/docs/fallbacks.rst
+++ b/docs/fallbacks.rst
@@ -1,13 +1,13 @@
 .. _fallbacks:
 
-Fallback / Default Actions
+Fallback Actions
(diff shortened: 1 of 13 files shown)


docs/fallbacks.rst
```diff
@@ -16,16 +16,21 @@ or if none of the dialogue policies predict an action with confidence higher tha
...
    agent = Agent("domain.yml",
                   policies=[KerasPolicy(), fallback])
 
 
-``action_listen`` will always be predicted after the fallback action, meaning
-the bot will wait for the next user input.
+``action_fallback`` is a default action in Rasa Core, which will send the
+``utter_default`` template message to the user. Make sure to specify this template
+in your domain file. It will also revert back to the state of the conversation
+before this action was executed, so that it will not influence the prediction of
```
tmbo: before the user message was received, right?

docs/fallbacks.rst
```diff
@@ -16,16 +16,21 @@ or if none of the dialogue policies predict an action with confidence higher tha
...
+future actions. You can take a look at the source of the action below:
```
tmbo: Let's add a note that you can also create your own action and use it as the fallback action.

docs/migrations.rst
```diff
@@ -5,6 +5,13 @@ Migration Guide
 This page contains information about changes between major versions and
 how you can migrate from one version to another.
 
+0.9.x to 0.10.0
+---------------
+
+- You can no longer load old models with this version, due to the addition of
```
tmbo: let's make this a `warning` as in the previous 0.8.x to 0.9.0

docs/fallbacks.rst
tmbo: and make sure the user knows, that they can not name their fallback action `action_fallback`

tmbo: probably better to name the default fallback `action_default_fallback` to avoid collisions with user actions

rasa_core/actions/action.py
```diff
@@ -105,3 +107,19 @@ def run(self, dispatcher, tracker, domain):
...
+    def name(self):
+        return ACTION_FALLBACK_NAME
+
+    def run(self, dispatcher, tracker, domain):
+        from rasa_core.events import UserUtteranceReverted, ActionReverted
+
+        if domain.random_template_for("utter_default") is not None:
+            dispatcher.utter_template("utter_default")
+
+        return [ActionReverted(), UserUtteranceReverted()]
```
tmbo: I don't think we need to revert the action first, if the utterance is reverted all actions after the utterance will be reverted anyways.

rasa_core/policies/fallback.py
```diff
@@ -92,10 +92,8 @@ def predict_action_probabilities(self, tracker, domain):
         # it is set to 1.0 here in order
         # to not override standard behaviour
         nlu_confidence = nlu_data["intent"].get("confidence", 1.0)
-        if tracker.latest_action_name == self.fallback_action_name:
```
tmbo: I thought about it, let's keep this. If the user creates his custom fallback action and doesn't do the user utterance reverted event, it would be good to keep this.

rasa_core/actions/action.py
akelad: oh nice, didn't know that

docs/fallbacks.rst
```diff
@@ -27,10 +27,17 @@ or if none of the dialogue policies predict an action with confidence higher tha
...
-.. autoclass:: rasa_core.actions.action.ActionFallback
+.. autoclass:: rasa_core.actions.action.ActionDefaultFallback
+
+.. note::
+  You can also create your own custom action to use as a fallback. Be aware
+  that if this action does not return a ``UserUtteranceReverted`` event, the
+  next predictions of your bot may become inaccurate, as it very likely that the
+  fallback action is not present in your stories
 
 If you have a specific intent that will trigger this, let's say it's called ``out_of_scope``, then you
```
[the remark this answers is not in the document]
akelad: Hmm I think the point that if you want a specific intent to trigger it is still valid though. Since this isn't actually the fallback policy that does this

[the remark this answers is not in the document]
tmbo: yes, your are right :+1:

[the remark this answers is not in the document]
akelad: cool, tests pass now too - are we good to merge?

PR #754, opened 2018-07-16
Fix removing generated OR checkpoint if it is the only one checkpoint

Ghostvv:
**Proposed changes**:
- remove redundant generated `OR` checkpoints even if there are no cycles in the stories
- remove the whole story step if such redundant generated `OR` checkpoints were the only checkpoints

- [ ] updated the changelog

### DIFF
--- a/data/test_stories/stories_checkpoint_after_or.md
+++ b/data/test_stories/stories_checkpoint_after_or.md
@@ -1,41 +1,9 @@
-## simple_story_without_checkpoint
-* simple                       <!-- user utterance in _intent[entities] format -->
-    - utter_default
-    - utter_greet
-
-## simple_story_with_only_start
-> check_greet                   <!-- checkpoints at the start define entry points -->
-* simple
-    - utter_default
-
-## simple_story_with_only_end
-* hello
-    - utter_greet
-    - slot{"name": "peter"}
-> check_greet                   <!-- checkpoint defining the end of this turn -->
-
-## simple_story_with_multiple_turns_or_only
+## story with checkpoint after or
 * affirm OR thank_you
 > check_after_or
 
-## simple_story_with_multiple_turns_after_or
+## story to continue checkpoint
 > check_after_or
     - utter_default
 * goodbye
     - utter_goodbye
-> check_goodbye        
-
-## why does the user want to leave?
-> check_goodbye
-* why
-    - utter_default
-> check_greet
-
-## show_it_all
-> check_greet
-> check_hello                   <!-- allows multiple entry points -->
-* next_intent
-    - utter_greet              <!-- actions taken by the bot -->
-> check_intermediate            <!-- allows intermediate checkpoints -->
-* change_bank_details
-    - utter_default            <!-- allows to end without checkpoints -->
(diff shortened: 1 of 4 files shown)


amn41:
@Ghostvv can you please add a test or 2?

Ghostvv:
it's a bit more complicated. I'll think about possible ones

rasa_core/training/structures.py
```diff
@@ -235,78 +235,76 @@ def with_cycles_removed(self):
         # type: () -> StoryGraph
         """Create a graph with the cyclic edges removed from this graph."""
 
-        if not self.cyclic_edge_ids:
-            return self
```
akelad: i don't think we should be getting rid of this?

Ghostvv: this is the main thing, `OR` checkpoints should be eliminated even if there are no cycles, so we need to call `_remove_unused_generated_cps`, so I inverted this `if` instead of calling `_remove_unused_generated_cps` two times

data/test_stories/stories_checkpoint_after_or.md
```diff
@@ -0,0 +1,41 @@
+## simple_story_without_checkpoint
+* simple                       <!-- user utterance in _intent[entities] format -->
+    - utter_default
+    - utter_greet
+
+## simple_story_with_only_start
```
tmbo: do we need all the other stories if we only want to test for the OR one?

rasa_core/training/structures.py
```diff
@@ -340,18 +338,29 @@ def _remove_unused_generated_cps(self, story_steps, overlapping_cps,
...
                            if cp_name.startswith(GENERATED_CHECKPOINT_PREFIX)}
 
+        k_to_remove = set()
         for k, step in story_steps.items():
             # changed all ends
             updated = step.create_copy(use_new_id=False)
             updated.start_checkpoints = self._checkpoint_difference(
                     updated.start_checkpoints, unused_overlapping_cps)
+            if step.start_checkpoints and not updated.start_checkpoints:
+                # remove story step if the generated checkpoints were unique
```
tmbo: I don't get this comment

rasa_core/training/structures.py
```diff
@@ -340,18 +338,29 @@ def _remove_unused_generated_cps(self, story_steps, overlapping_cps,
...
+                k_to_remove.add(k)
 
             # remove generated unused end checkpoints
             updated.end_checkpoints = self._checkpoint_difference(
                     updated.end_checkpoints, unused_genr_cps)
+            if step.end_checkpoints and not updated.end_checkpoints:
```
tmbo: i think there is no reason to split the if,

```
            if (step.end_checkpoints and not updated.end_checkpoints or 
                    step.start_checkpoints and not updated.start_checkpoints):
                k_to_remove.add(k)

```

rasa_core/training/structures.py
Ghostvv: I meant  `remove story step if the generated checkpoints were the only ones`

tmbo: :+1:

PR #767, opened 2018-07-19
Specify YAML version 1.1 for ruamel parser

### LINKED ISSUE #765: Buttons with Payloads / Titles containing "Yes" or "No" get transformed incorrectly
**Rasa Core version**: 0.10.1

**Python version**: 3.6

**Operating system** (windows, osx, ...): Ubuntu

**Issue**:

After upgrading from 0.9 to 0.10.1 I noticed the following phenomenon when running my unit tests: a button utterance defined like this:

```yml
  utter_buttons:
  - text: "Buttons, there will be buttons:"
    buttons:
      - title: "Yes"
        payload: "Yes"
      - title: "No"
        payload: "No"
```

Will be serialized like this when using a `CollectingOutputChannel`:

```json
{
    "messages": [
        {
            "data": [
                {
                    "payload": true,
                    "title": true
                },
                {
                    "payload": false,
                    "title": false
                }
            ],
            "recipient_id": "xxx",
            "text": "Buttons, there will be buttons:"
        }
    ]
}
```

when using 0.9 would do the expected:

```json
{
    "messages": [
        {
            "data": [
                {
                    "payload": "Yes",
                    "title": "Yes"
                },
                {
                    "payload": "No",
                    "title": "No"
                }
            ],
            "recipient_id": "xxx",
            "text": "Buttons, there will be buttons:"
        }
    ]
}
```

Any string that is not **"Yes"** or **"No"** will still work as expected in 0.10.1

m90:
**Proposed changes**:

In order to normalize python2 / python3 behaviors, make
ruamel.yaml use YAML 1.1 like pyyaml does. Fixes #765.

- [ ] updated the changelog

I'm still not sure why `ruamel.yaml` creates invalid YAML on unmarshaling / marshaling, but since Python 2 users will have to deal with YAML 1.1 anyways I think this would be the best solution anyways.


rasa_core/utils.py
```diff
@@ -320,7 +320,7 @@ def read_yaml_string(string):
         import ruamel.yaml
 
         yaml_parser = ruamel.yaml.YAML(typ="safe")
-        yaml_parser.allow_unicode = True
```
m90: `allow_unicode` defaults to `True`, so I removed it.

tests/test_domain.py
```diff
@@ -184,5 +184,5 @@ def test_domain_to_yaml():
   utter_greet:
   - text: hey there!"""
     domain = TemplateDomain.load_from_yaml(test_yaml)
-    assert test_yaml.strip() == domain.as_yaml().strip()
+    assert domain.as_yaml().strip().endswith(test_yaml.strip())
```
m90: Using Python 3 the document will start with a pragma when it won't with Python 2, so `==` won't work anymore.

akelad:
Just had a look at the related issue, very strange. Let me have a look first what changed from 0.9 to 0.10 that is causing this and then I'll review this PR. Thanks for taking the time to investigate this issue and create a PR for it!

tests/test_domain.py
tmbo: do you mind adding this as a comment so we keep this in mind?

PR #776, opened 2018-07-20
New chat input channel

ricwo:
**Proposed changes**:
- add ``RasaChatInput``

- [x] updated the changelog

### DIFF
--- a/rasa_core/channels/rasa_chat.py
+++ b/rasa_core/channels/rasa_chat.py
@@ -21,12 +21,40 @@ class NoOutputChannel(OutputChannel):
     def send_text_message(self, recipient_id, message):
         # type: (Text, Text) -> None
         """Just ignore everything."""
+
         pass
 
 
 class RasaChatInput(HttpInputComponent):
+    """Chat input channel for Rasa Platform"""
+
     def __init__(self, host):
         self.host = host
+
+    def _check_token(self, token):
+        url = "{}/users/me".format(self.host)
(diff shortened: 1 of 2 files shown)


rasa_core/channels/rasa_chat.py
```diff
@@ -0,0 +1,82 @@
...
+from flask import Blueprint, request, jsonify, abort
+from flask_cors import CORS, cross_origin
+
+from rasa_core.channels.channel import UserMessage, OutputChannel
+from rasa_core.channels.rest import HttpInputComponent
+
+logger = logging.getLogger(__name__)
+
+
+class NoOutputChannel(OutputChannel):
```
amn41: would be just use the `CollectingOutputChannel` for this? https://github.com/RasaHQ/rasa_core/blob/master/rasa_core/channels/direct.py#L4

rasa_core/channels/rasa_chat.py
```diff
@@ -0,0 +1,82 @@
...
+        @rasa_chat.route("/", methods=['GET'])
+        def health():
+            return jsonify({"status": "ok"})
+
+        @rasa_chat.route("/send", methods=['GET', 'POST'])
+        @cross_origin()
+        def receive():
+            user = self.fetch_user(request)
+            msg = request.json["message"]
+            if user.get("role") == "admin":
```
amn41: I don't think we need this - all users should be treated the same regardless of role

PR #791, opened 2018-07-23
Unfeaturized entities for specific intents

akelad:
**Proposed changes**:
- allow user to specify a list of intents in the domain file for which entities should be ignored. These entities will only be ignored if they do not have a featurized slot of the same name.

- [x] updated the changelog

### DIFF
--- a/docs/domains.rst
+++ b/docs/domains.rst
@@ -5,8 +5,10 @@ Domain Format
 
 The ``Domain`` defines the universe in which your bot operates.
 It specifies the ``intents``, ``entities``, ``slots``, and ``actions``
-your bot should know about. 
+your bot should know about.
 Optionally, it can also include ``templates`` for the things your bot can say.
+You can also define a list of intents, under ``intents_ignore_entities``, for
+which Rasa Core should ignore any entities found. By default no entities will be ignored.
 
 
 As an example, the ``DefaultDomain`` has the following yaml definition:
@@ -37,7 +39,7 @@ Custom Actions and Slots
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
 To reference custom actions and slots in your domain,
-you need to reference them by their module path. 
+you need to reference them by their module path.
 For example, if you have a module called ``my_actions`` containing
 a class ``MyAwesomeAction``, and module ``my_slots`` containing ``MyAwesomeSlot``,
 you would add these lines to the domain file:
@@ -143,4 +145,15 @@ multiple responses and Rasa will randomly pick one of them, e.g.:
     - text: "Hey, {name}. How are you?"
     - text: "Hey, {name}. How is your day going?"
 
+Ignoring entities for certain intents
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+If you want entities to be ignored for certain intents, list the intents under
+``intents_ignore_entities``. This means that if the entities for those intents
+don't have a featurized slot associated with them, Rasa Core will unfeaturize
+them and therefore will not impact its predictions. This is useful when you have
+an intent where you don't care about the entities being picked up.
+
+.. note::
+    If your slot of the same name as an entity does not have a type ``unfeaturized``
+    that entity will not be unfeaturized.
--- a/docs/domains.rst
+++ b/docs/domains.rst
@@ -7,8 +7,6 @@ The ``Domain`` defines the universe in which your bot operates.
 It specifies the ``intents``, ``entities``, ``slots``, and ``actions``
 your bot should know about.
 Optionally, it can also include ``templates`` for the things your bot can say.
-You can also define a list of intents, under ``intents_ignore_entities``, for
-which Rasa Core should ignore any entities found. By default no entities will be ignored.
 
 
 As an example, the ``DefaultDomain`` has the following yaml definition:
@@ -148,12 +146,21 @@ multiple responses and Rasa will randomly pick one of them, e.g.:
 Ignoring entities for certain intents
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-If you want entities to be ignored for certain intents, list the intents under
-``intents_ignore_entities``. This means that if the entities for those intents
-don't have a featurized slot associated with them, Rasa Core will unfeaturize
-them and therefore will not impact its predictions. This is useful when you have
-an intent where you don't care about the entities being picked up.
+If you want entities to be ignored for certain intents, you can add the ``use_entities: false``
+parameter to the intent in your domain file like this:
+
+.. code-block:: yaml
+
+  intents:
+    - greet:
+      use_entities: false
+
+This means that entities for those intents will be unfeaturized and therefore
+will not impact the next action predictions. This is useful when you have
+an intent where you don't care about the entities being picked up. If you list
+your intents as normal without this flag, the entities will be featurized as normal.
 
 .. note::
-    If your slot of the same name as an entity does not have a type ``unfeaturized``
-    that entity will not be unfeaturized.
+
+    If you really want these entities not to influence action prediction we
+    suggest you make the slots with the same name of type ``unfeaturized``
--- a/rasa_core/domain.py
+++ b/rasa_core/domain.py
@@ -232,13 +232,13 @@ class Domain(with_metaclass(abc.ABCMeta, object)):
         # Set all found entities with the state value 1.0
         for entity in tracker.latest_message.entities:
             entity_name = entity.get("entity")
-            should_ignore_entities = (tracker.latest_message.intent.get("name")
-                                      in self.intents_ignore_entities)
+            should_ignore_entity = (tracker.latest_message.intent.get("name")
+                                    in self.intents_ignore_entities)
             entity_is_featurized_slot = (entity_name in tracker.slots.keys()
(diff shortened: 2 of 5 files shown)


rasa_core/domain.py
```diff
@@ -224,16 +224,25 @@ def input_states(self):
...
-            key = "entity_{0}".format(entity["entity"])
-            state_dict[key] = 1.0
+            entity_name = entity.get("entity")
+            should_ignore_entity = (tracker.latest_message.intent.get("name")
+                                    in self.intents_ignore_entities)
+            entity_is_featurized_slot = (entity_name in tracker.slots.keys()
+                                         and tracker.slots[entity_name].type_name
+                                         != 'unfeaturized')
+            if (not should_ignore_entity or (should_ignore_entity and
+                                             entity_is_featurized_slot)):
```
Ghostvv: `if not should_ignore_entity or  entity_is_featurized_slot:` is the same, no?

akelad: yes, finally a better or statement 😄

docs/domains.rst
```diff
@@ -5,8 +5,10 @@ Domain Format
 
 The ``Domain`` defines the universe in which your bot operates.
 It specifies the ``intents``, ``entities``, ``slots``, and ``actions``
-your bot should know about. 
+your bot should know about.
 Optionally, it can also include ``templates`` for the things your bot can say.
+You can also define a list of intents, under ``intents_ignore_entities``, for
```
Ghostvv: I think we should think of a better name for `intents_ignore_entities`

akelad: yeah i was also thinking that, any suggestions? i couldn't come up with anything good sounding that wasn't also reeeeally long

Ghostvv: yeah, me too :grinning:

rasa_core/domain.py
```diff
@@ -224,16 +224,24 @@ def input_states(self):
...
         for entity in tracker.latest_message.entities:
-            key = "entity_{0}".format(entity["entity"])
-            state_dict[key] = 1.0
+            entity_name = entity.get("entity")
+            should_ignore_entity = (tracker.latest_message.intent.get("name")
+                                    in self.intents_ignore_entities)
+            entity_is_featurized_slot = (entity_name in tracker.slots.keys()
+                                         and tracker.slots[entity_name].type_name
+                                         != 'unfeaturized')
+            if not should_ignore_entity or entity_is_featurized_slot:
```
tmbo: is this the desired behaviour? it ignores all entities if they are not mapped to a slot?

akelad: The original statement was this: `if (not should_ignore_entity or (should_ignore_entity and entity_is_featurized_slot))`, but that one equates to the same thing. So ignore the entities only if they're part of an intent in the list and not a featurized slot

tests/test_domain.py
```diff
@@ -179,6 +179,7 @@ def test_domain_to_yaml():
   store_entities_as_slots: true
 entities: []
 intents: []
+intents_ignore_entities: []
```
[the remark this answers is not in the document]
akelad: you mean the intents array?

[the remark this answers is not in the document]
tmbo: yes, so making the elements of that array of type `union[text, dict]`

rasa_core/domain.py
```diff
@@ -224,16 +224,24 @@ def input_states(self):
...
 
-        # Set all found entities with the state value 1.0
+        # Set all found entities with the state value 1.0, unless they should
+        # be ignored for the current intent and are not associated with a
+        # featurized slot
         for entity in tracker.latest_message.entities:
-            key = "entity_{0}".format(entity["entity"])
-            state_dict[key] = 1.0
+            entity_name = entity.get("entity")
+            should_ignore_entity = (tracker.latest_message.intent.get("name")
```
tmbo: can we name this in a positive way and invert it, bools are easier to read if their meaning is positive.

`should_use_entity = (tracker.latest_message.intent.get("name") not in self.intents_ignore_entities)`

rasa_core/domain.py
tmbo: thanks :+1: why do we featurize the entity if it maps to a slot and is ignored? won't the slot be featurized and there is essentially no need to featurize the entity as well?

akelad: good point, i think it was just to keep it how we do it currently. @JoeyFaulkner @Ghostvv was there a particular reason we decided to do it this way?

Ghostvv: We need to featurize the entity for the cases like `correct{"price": "expensive"}`, in this case the slot was set before, but we need to react with certain entity to the fact that it was input again

tmbo: but why would you ignore the entity in this case in the first place? or should we rather have something like:
```
intents:
- greet:
   use_entities: ["name"]
```
meaning the intent will ignore all entities but the name?

akelad: yeah i agree, in that case you shouldn't unfeaturize entities @Ghostvv . but i'm not sure we need your suggestion @tmbo - i don't see a case where you'd want to pay attention to one but not the rest?

tmbo: but isn't the example @Ghostvv posted an example for this - ignore all the entities, but the "price" entity?

akelad: no because in the `correct` case you want to pay attention to all entities, hence why I said you shouldn't ignore them in the first place

tests/test_domain.py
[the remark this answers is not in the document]
akelad: hmm i'm just thinking how should this look like ideally, `["intent_greet",  {"intent_goodbye": false}]` - like this if `greet` should pay attention to entities and `goodbye` shouldn't or how?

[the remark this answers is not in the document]
tmbo: you mean after parsing the yaml?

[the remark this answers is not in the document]
akelad: yes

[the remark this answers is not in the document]
tmbo: after parsing it, the format should be consistent for all intents, probably something like:
```
{"intent_greet": {"use_entities": True}, "intent_goodbye": {"use_entities": False}}
```
so basically a dict with the intents as keys and properties of them as values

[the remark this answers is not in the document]
akelad: that's not going to be backwards compatible then though is it?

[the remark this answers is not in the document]
tmbo: but this is just about the internal representation inside the templateddomain class right? as long as the code can still read the old yaml domain yaml files and does the same thing, we are good.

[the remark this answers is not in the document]
akelad: yes true, never mind that point :P

rasa_core/domain.py
tmbo: ok in that case and if there are no objections I'd suggest to:
- not featurize ignored entities, even if there are slots
- keep the simple boolean flag, we can still later extend our `"use_entities": false` to `"use_entities": ["name"]` later if we need it

akelad: @Ghostvv no we're not defeaturizing slots, we're just not featurizing the entities even if their slots are featurized. So like this: 
```
            should_use_entity = (tracker.latest_message.intent.get("name") not
                                 in self.intents_ignore_entities)
            if should_use_entity:
                key = "entity_{0}".format(entity["entity"])
                state_dict[key] = 1.0
```

Ghostvv: but we already discussed it? it will cause different feature vectors during prediction and train times if train stories were without entities and, it will cause the same branching if entities are still present in train story. I agree that resulting behavior is the same, but this way we create somewhat hidden behavior that can be difficult to understand and debug for users

akelad: i'm not entirely sure you're thinking of the same thing as us? of course there will still be tracker branching, but it's going to do that anways since they're associated with featurized slots

Ghostvv: ok, I think i get now

akelad: this is super hard to discuss via github 🙈

akelad:
ok @tmbo this is ready for another review (apart from the docs)

rasa_core/domain.py
```diff
@@ -507,10 +521,9 @@ def as_dict(self):
         additional_config = {
             "store_entities_as_slots": self.store_entities_as_slots}
         action_names = self.action_names[len(Domain.DEFAULT_ACTIONS):]
-
         domain_data = {
             "config": additional_config,
-            "intents": self.intents,
+            "intents": [{k: v} for k, v in self._intents.items()],
```
tmbo: i think `list(self._intents.items())` should do the same

rasa_core/schemas/domain.yml
```diff
@@ -3,6 +3,8 @@ mapping:
   intents:
     type: !!python/str "seq"
     sequence:
+    - type: !!python/str "map"
+      allowempty: True
```
tmbo: let's also add the `mapping: ...` for the `use_entities` here as well

rasa_core/domain.py
tmbo: ah right, for backward compatibility, this should just be the list of intents

tmbo: no that won't work, you are right

tmbo: ok ✅

PR #826, opened 2018-08-01
add respond to remote core

tmbo:
**Proposed changes**:
- added respond to remote core

- [ ] updated the changelog

### DIFF
--- a/rasa_core/remote.py
+++ b/rasa_core/remote.py
@@ -125,6 +125,28 @@ class RasaCoreClient(object):
                                               response.text))
             return None
 
+    def respond(self, sender_id, message):
+        # type: (Text, Text) -> Optional[Dict[Text, Any]]
+        """Send a parse request to a rasa core server."""
+
+        url = "/conversations/{}/respond".format(sender_id)
+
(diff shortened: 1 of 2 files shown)


rasa_core/remote.py
```diff
@@ -125,6 +125,28 @@ def parse(self, message, sender_id):
                                               response.text))
             return None
 
+    def respond(self, sender_id, message):
+        # type: (Text, Text) -> Optional[Dict[Text, Any]]
+        """Send a parse request to a rasa core server."""
+
+        url = "/conversations/{}/respond".format(sender_id)
```
Ghostvv: seems to be a lot of duplicating code with `parse(..)` except for this line, should it be one function with request type parameter?

Ghostvv: and `continue_core` seems to be the same except for `data`

tmbo:
Thanks :+1: you are completely on spot with the code duplication. I'd like to keep it this way though, as it makes the code easier to read / maintain - the requests code is boilerplate anyways and doesn't contain any logic. that the requests to the endpoints are very similar is more coincidental than expected :D

PR #846, opened 2018-08-06
update nlu req

amn41:
**Proposed changes**:
- ...

- [ ] updated the changelog


tmbo:
why does binder have it's own requirements file in the first place, can't it just use the requirements file we already have?

amn41:
I seem to remember it was needed for some reason, but let's try without and check it works

amn41:
yep also works with a simple include 💯

PR #851, opened 2018-08-08
Predict endpoint clean

JoeyFaulkner:
**Proposed changes**:
- add /predict endpoint and endpoint for the time the agent was last trained

- [ ] updated the changelog

### DIFF
--- a/rasa_core/server.py
+++ b/rasa_core/server.py
@@ -23,6 +23,7 @@ from rasa_core.channels.direct import CollectingOutputChannel
 from rasa_core.tracker_store import TrackerStore
 from rasa_core.trackers import DialogueStateTracker
 from rasa_core.version import __version__
+from rasa_core.channels import UserMessage
 
 if typing.TYPE_CHECKING:
     from rasa_core.interpreter import NaturalLanguageInterpreter as NLI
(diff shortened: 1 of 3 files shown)


rasa_core/server.py
```diff
@@ -432,6 +451,13 @@ def load_model():
         logger.debug("Finished loading new agent.")
         return jsonify({'success': 1})
 
+    @app.route("/model/time-of-last-train", methods=['GET'])
+    @requires_auth(auth_token)
```
tmbo: This should rather use a property dumped to the metadata when persisted. The model might not habe been trained on this rasa core instance

JoeyFaulkner: well the property is on the agent, so should i just include this property when persisting the agent?

tmbo: The issue is that if you load a persisted model, the time-of-last-train will always be None, even if you load a different model.

My idea was, to persist a uuid, timestamp, ... as part of the metadata https://github.com/RasaHQ/rasa_core/blob/master/rasa_core/policies/ensemble.py#L139 so we can use it when loading the model.

JoeyFaulkner: that make sense!

tmbo:
do you mind rebasing this onto the `dispatcher` branch and using `agent.fingerprint` instead of the `time_of_last_train`?

PR #872, opened 2018-08-14
Removing CMD statement from Dockerfile

tijsg:
When using this docker component in a cloud environment, the dialogue model and nlu model might still have to be uploaded to the pod. The starting of the docker container should not depend on there already being models present inside the container.

**Proposed changes**:
- ...

- [ ] updated the changelog


akelad:
Thanks for your contribution -- we're planning on changing the Dockerfile anyways to only load up the core model and then communicate with an NLU container from there. We're currently doing a big SDK refactor of Core, so once this is done we'll make sure we provide appropriate Dockerfiles afterwards too. So I'd say we close this PR, unless you have any objections @tmbo ?

tijsg:
Ok thanks for the feedback!
actually this one seems to work fine:

FROM python:3.6.6-slim
RUN apt-get update -y
RUN apt-get install build-essential -y
RUN pip install rasa_core

can be added as automated build to docker hub, to have automatic upgrades of the images

tijsg:
Please check out this one:

https://hub.docker.com/r/tijsg/rasa-core/

PR #943, opened 2018-09-05
Fallback update

### LINKED ISSUE #749: Trigger fallback policy when ActionListen is predicted after a user utterance
Because generally the most commonly predicted action is ActionListen, when the bot is out of its comfort zone it may just predict ActionListen after a user utterance, which doesn't help the conversation move along. It should go to fallback policy in this case.
Case when this should happen:
1. if there’s a fallbackpolicy present AND
2. if there was just a user message AND
3. the predicted action is action_listen AND
4. it wasn't predicted by the memo policy

### LINKED ISSUE #800: Bug in SimplePolicyEnsemble probability normalization
**Rasa Core version**:
master

**Python version**: 
3.6

Ubuntu

**Issue**:

I believe there is a bug in this line: https://github.com/RasaHQ/rasa_core/blob/e5a5404916ebaae1b425b3d6f20a191c5727caa8/rasa_core/policies/ensemble.py#L223

The intent is to have the probabilities sum to 1 but this is not what `np.linalg.norm` does by default. A better solution would be:

`result = result / np.sum(result)`

This is especially problematic for lines like this:

https://github.com/RasaHQ/rasa_core/blob/master/rasa_core/policies/fallback.py#L105

since dividing by L2-norm can lead to probabilities > 1.

Furthermore, in the [line above](/blob/e5a5404916ebaae1b425b3d6f20a191c5727caa8/rasa_core/policies/ensemble.py#L222), comparing floats with `==` is not advisable, better use `np.isclose`.

akelad:
**Proposed changes**:
- predict fallback policy on action listen after user utterance (provided it wasn't predicted by the memo policy)
- fixes #749 
- fixes #800 

- [ ] updated the changelog


rasa_core/policies/ensemble.py
```diff
@@ -208,9 +208,29 @@ def probabilities_using_best_policy(self, tracker, domain):
...
                 best_policy_name = 'policy_{}_{}'.format(i, type(p).__name__)
+
+        policy_names = [type(p).__name__ for p in self.policies]
+
+        # Trigger the fallback policy when ActionListen is predicted after
+        # a user utterance. This is done on the condition that: a fallback
+        # policy is present, there was just a user message and the predicted
+        # action is action_listen by a policy other than the MemoizationPolicy
+
+        if "FallbackPolicy" in policy_names:
```
tmbo: this breaks if we ever rename this, let's make it fail in that case:
`FallbackPolicy.__name__ in policy_names`

rasa_core/policies/ensemble.py
```diff
@@ -208,9 +208,29 @@ def probabilities_using_best_policy(self, tracker, domain):
...
+            idx = policy_names.index("FallbackPolicy")
+            fallback_policy = self.policies[idx]
+
+            if (result.index(max_confidence) == 0 and
+                not best_policy_name.endswith("MemoizationPolicy") and
```
tmbo: same with `MemoizationPolicy `

tmbo: what if this is the augmented memoization policy?

tmbo: let's make that explicit:
```
best_policy_name.endswith("." + MemoizationPolicy.__name__) or best_policy_name.endswith("." + AugmentedMemoizationPolicy.__name__)
``` 
and probably best to put the whole if condition in a separate function

rasa_core/policies/ensemble.py
```diff
@@ -208,9 +208,29 @@ def probabilities_using_best_policy(self, tracker, domain):
...
+                    isinstance(tracker.events[-1], UserUttered)):
+                result = fallback_policy.predict_action_probabilities(
+                                            tracker, domain, force=True)
+                best_policy_name = 'policy_{}_{}'.format(idx, "FallbackPolicy")
```
tmbo: `type(fallback_policy).__name__`

rasa_core/policies/ensemble.py
```diff
@@ -208,9 +208,29 @@ def probabilities_using_best_policy(self, tracker, domain):
...
+
         # normalize probablilities
         if np.sum(result) != 0:
-            result = result / np.linalg.norm(result)
+            result = result / np.nansum(result)
+            print(result)
```
tmbo: please remove / replace

rasa_core/policies/ensemble.py
```diff
@@ -208,9 +208,29 @@ def probabilities_using_best_policy(self, tracker, domain):
...
         logger.debug("Predicted next action using {}"
                      "".format(best_policy_name))
-        return result
+        return result, best_policy_name
```
tmbo: type annotation of function needs to change to `Tuple[List[float], Text]`

rasa_core/policies/fallback.py
```diff
@@ -92,8 +92,12 @@ def predict_action_probabilities(self, tracker, domain):
         # it is set to 1.0 here in order
         # to not override standard behaviour
         nlu_confidence = nlu_data["intent"].get("confidence", 1.0)
-
-        if tracker.latest_action_name == self.fallback_action_name:
+        if force:
+            logger.debug("Action listen was predicted after a user message. "
+                         "Predicting fallback action: {}"
+                         "".format(self.fallback_action_name))
+            score = 1.1
```
tmbo: lets move the `result` creation into a separate function

```python
def fallback_scores(self, domain, fallback_score):
    """Prediction scores used if a fallback is necessary."""
    
    result = [0.0] * domain.num_actions
    idx = domain.index_for_action(self.fallback_action_name)
    result[idx] = fallback_score
    return result
```
and then we can use that function in all the existing if branches here and instead of using the `force` parameter we do not call `predict_action_probabilities` from the policy ensemble in the action listen fallback case, but instead we use the above `fallback_scores`. Feel free to also make the `1.1` a constant so we can use it from the ensemble.

rasa_core/processor.py
```diff
@@ -98,7 +98,7 @@ def predict_next(self, sender_id):
                            "'{}'.".format(sender_id))
             return None
 
-        probabilities = self._get_next_action_probabilities(tracker)
+        probabilities, _ = self._get_next_action_probabilities(tracker)
```
tmbo: let's add the best policy info to the returned object as well

rasa_core/processor.py
```diff
@@ -150,13 +150,13 @@ def predict_next_action(self, tracker):
...
 
-        probabilities = self._get_next_action_probabilities(tracker)
+        probabilities, policy = self._get_next_action_probabilities(tracker)
 
         max_index = int(np.argmax(probabilities))
         action = self.domain.action_for_index(max_index, self.action_endpoint)
         logger.debug("Predicted next action '{}' with prob {:.2f}.".format(
                 action.name(), probabilities[max_index]))
-        return action
+        return action, policy, probabilities[max_index]
```
tmbo: type annotation needs to change

rasa_core/policies/ensemble.py
akelad: I mean they both end with "MemoizationPolicy" :D but yeah this makes sense if we rename it in future

akelad: also do you mean the whole if statement or just the augmentedmemo/memoization policy one?

tmbo: whatever you think works better, the if should just be easier to read

akelad: ok i think i'll just do the augmented/memo policy one

akelad:
@tmbo this is ready for another review now

PR #958, opened 2018-09-07
Consider port overrides when constructing server URLs

m90:
**Proposed changes**:

I found that when using the `cmdline` channel, a possible port override passed to `serve_application` will not be considered as the hardcoded URL that is already formatted with the default port is being passed here: https://github.com/RasaHQ/rasa_core/blob/a26dd755492a0c3d3516229967b38bf8930e8179/rasa_core/run.py#L209

creating a possible mismatch between listener and caller. Instead, always consider possible port overrides when passing the server URL.

- [ ] updated the changelog


tmbo:
Thanks for the fix :+1:

PR #970, opened 2018-09-10
Cli more flags

amn41:
**Proposed changes**:
- user can choose if they want to finetune the model after giving feedback (deafult False)
- user can provide a model to load instead of pre-training from data. This should probably be made a mutually exclusive group with all the training-related cmdline args
- add exit code flag to evaluate.py

- [ ] updated the changelog


rasa_core/evaluate.py
```diff
@@ -60,7 +60,11 @@ def create_argument_parser():
             type=str,
             default="failed_stories.txt",
             help="output path for the failed stories")
-
+    parser.add_argument(
```
tmbo: this has already been added in #966 as `--fail_on_prediction_errors `

rasa_core/train.py
```diff
@@ -82,6 +82,15 @@ def create_argument_parser():
             default=False,
             action='store_true',
             help="enable online training")
+    parser.add_argument(
```
tmbo: should we use the same name as we use in `run`?  `--core`

rasa_core/training/online.py
```diff
@@ -469,8 +469,8 @@ def _write_stories_to_file(export_file_path, sender_id, endpoint):
 
 
 def _predict_till_next_listen(endpoint,  # type: EndpointConfig
-                              sender_id  # type: Text
-                              ):
+                              sender_id,  # type: Text
+                              finetune): # type: Boolean
```
tmbo: type needs to come directly after the argument (`):`) needs to go onto the next line

rasa_core/evaluate.py
amn41: 👌

amn41:
thanks @tmbo for comments. I just log now that all training params are ignored if loading a pre-trained model

rasa_core/train.py
```diff
@@ -30,7 +30,7 @@ def create_argument_parser():
 
     # either the user can pass in a story file, or the data will get
     # downloaded from a url
-    group = parser.add_mutually_exclusive_group(required=True)
+    group = parser.add_mutually_exclusive_group(required=False)
```
tmbo: how about keeping the exclusive group with required and adding the `--core` to that group - so either you need to supply training data or a model

amn41:
cool @tmbo please have another look

PR #987, opened 2018-09-12
added socketio channel implementation

tmbo:
**Proposed changes**:
- added socketio channel implementation

- [x] updated the changelog

### DIFF
--- a/rasa_core/channels/socketio.py
+++ b/rasa_core/channels/socketio.py
@@ -130,7 +130,7 @@ class SocketIOInput(InputChannel):
         def connect(sid, environ):
             logger.debug("User {} connected to socketio endpoint.".format(sid))
 
-        @sio.on('disconnect', namespace=self.name())
+        @sio.on('disconnect', namespace=self.namespace)
         def disconnect(sid):
             logger.debug("User {} disconnected from socketio endpoint."
                          "".format(sid))
--- a/rasa_core/channels/socketio.py
+++ b/rasa_core/channels/socketio.py
@@ -1,0 +1,144 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import logging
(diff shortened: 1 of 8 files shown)


rasa_core/channels/socketio.py
```diff
@@ -0,0 +1,144 @@
...
+    def from_credentials(cls, credentials):
+        credentials = credentials or {}
+        return cls(credentials.get("user_message_evt", "user_uttered"),
+                   credentials.get("bot_message_evt", "bot_uttered"),
+                   credentials.get("namespace"))
+
+    def __init__(self,
+                 user_message_evt="user_uttered",  # type: Text
+                 bot_message_evt="bot_uttered",  # type: Text
+                 namespace=None  # type: Optional[Text]
```
akelad: what's this namespace for? if we're reading it from the credentials file it should also be explained in the docs right?

rasa_core/channels/socketio.py
```diff
@@ -0,0 +1,144 @@
...
+
+        @socketio_webhook.route("/", methods=['GET'])
+        def health():
+            return jsonify({"status": "ok"})
+
+        @sio.on('connect', namespace=self.namespace)
+        def connect(sid, environ):
+            logger.debug("User {} connected to socketio endpoint.".format(sid))
+
+        @sio.on('disconnect', namespace=self.name())
```
akelad: why is this `self.name()` and the `connect` one  above `self.namespace`?

rasa_core/channels/socketio.py
tmbo: no clue :D didn't read the socket io docs - but i think it allows you to use the same message names in different contexts. don't think it is important though.

rasa_core/channels/socketio.py
tmbo: good spot, should be namespace as well :+1:

PR #989, opened 2018-09-12
fixes #988: fix fallback for two action_listen in a row

### LINKED ISSUE #988: Fallback for ActionListen doesn't work if a slot was set by the user utterance
Basically, this `if isinstance(tracker.events[-1], UserUttered)` isn't true if a slot was set by the user utterance. 
https://github.com/RasaHQ/rasa_core/blob/master/rasa_core/policies/ensemble.py#L232

Ghostvv:
**Proposed changes**:
- check if latest action is action_listen instead of checking if last event is UserUttered
- fixes #988

- [ ] updated the changelog


rasa_core/policies/ensemble.py
```diff
@@ -198,10 +200,10 @@ def continue_training(self, trackers, domain, **kwargs):
 
 class SimplePolicyEnsemble(PolicyEnsemble):
 
-    def is_not_memo_policy(self, best_policy_name):
-        return not (best_policy_name.endswith("_" + MemoizationPolicy.__name__)
-                    or best_policy_name.endswith(
-                                    "_" + AugmentedMemoizationPolicy.__name__))
+    @staticmethod
+    def is_not_memo_policy(best_policy_name):
+        # AugmentedMemoizationPolicy also ends with MemoizationPolicy
```
akelad: yep, but we might change this, so please put it back to the way it was

rasa_core/policies/ensemble.py
```diff
@@ -216,27 +218,32 @@ def probabilities_using_best_policy(self, tracker, domain):
                 result = probabilities
                 best_policy_name = 'policy_{}_{}'.format(i, type(p).__name__)
 
-        policy_names = [type(p).__name__ for p in self.policies]
-
-        # Trigger the fallback policy when ActionListen is predicted after
-        # a user utterance. This is done on the condition that: a fallback
-        # policy is present, there was just a user message and the predicted
-        # action is action_listen by a policy other than the MemoizationPolicy
+        if (result.index(max_confidence) ==
```
akelad: I think it makes more sense to check if a fallback policy is present first, because otherwise we don't need to check for all this

rasa_core/policies/ensemble.py
```diff
@@ -13,15 +13,17 @@
 import numpy as np
 from builtins import str
 import typing
-from typing import Text, Optional, Any, List, Dict
+from typing import Text, Optional, Any, List, Dict, Tuple
```
akelad: why are we importing the Tuple? It's not used anywhere

Ghostvv: it is used here: https://github.com/RasaHQ/rasa_core/blob/b9fa9963709bb86d2bb9fa19b8605c6b99227e0e/rasa_core/policies/ensemble.py#L209

akelad: ahh got it, didn't see that 😄

rasa_core/policies/ensemble.py
Ghostvv: I'm not sure here. Then we need to iterate through `policies` arrays to find fallback policy every time. And I think, this way it is easier for users to override this behavior

akelad: haha since we decided to make it specific to the fallback policy, it is fallback policy specific 😄

tmbo: I think either way is fine :+1:

Ghostvv: to me, if I were a user and I wanted to create specific policy for this, this would be easier to change

akelad: 🤷‍♀️ i still think it makes more sense the other way, but whatever, this PR needs to be merged so lets merge it

Ghostvv: dark side of the force clouds your vision, come to the light side 😂

PR #994, opened 2018-09-13
Daemonize threaded processes used for console I/O

m90:
**Proposed changes**:
If you wrap `run.serve_application`  or `training.online.serve_application` in a custom CLI tool (or any other kind of application) and use console I/O, this application will currently not exit cleanly on `SIGINT` as the child processes summoned by rasa are not running as daemons and at the same time noone is listening to any signals in order to shut them down manually.

This means it currently requires two SIGINTs for the parent application to exit (which unfortunately apparently also kills/freezes Conda on Windows on the second signal).

To fix this condition, have the child processes run daemonized (as already done in  `agent.start_model_pulling_in_worker`).

- [ ] updated the changelog

### DIFF
--- a/rasa_core/run.py
+++ b/rasa_core/run.py
@@ -156,6 +156,7 @@ def start_cmdline_io(server_url, on_finish, **kwargs):
 
     p = Thread(target=console.record_messages,
                kwargs=kwargs)
+    p.setDaemon(True)
     p.start()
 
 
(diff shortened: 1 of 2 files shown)


akelad:
Thanks a lot for submitting this pull request, we'll give it a review as soon as possible

tmbo:
Very good idea, thanks 👍

PR #1002, opened 2018-09-17
fix branch whitelist

amn41:
**Proposed changes**:
- don't match forked master branches

- [ ] updated the changelog

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -356,8 +356,7 @@ scv_show_banner = True
 scv_banner_greatest_tag = True
 scv_sort = ('semver',)
 scv_overflow = ("-A", "html_theme=rasabaster")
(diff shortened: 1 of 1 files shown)


tmbo:
but does this really prevent forked masters? - I though the travis build configuration does that by not building the docs

amn41:
well previously we were matching things like `username-master` which I presume are forks. spinxcontrib-versioning pulls every ref from remote before checking against the whitelist, so it's not so much about what travis builds but what's in the git remote

tmbo:
right, I wasn't aware that the plugin is treating the `svc_whitelist_branches` as `.*master.*` - thought it would just do a `==`.

amn41:
yep was also a surprise to me :)

Issue #1035, opened 2018-09-20
tracker_store can be added to endpoints.yml


souvikg10:
**Rasa Core version**:0.11.3

**Python version**: 3.6

**Operating system** (windows, osx, ...): linux

**Issue**:

I think it would be interesting if we can configure the tracker_store info in the endpoints.yml file

```
tracker_store:
 type: redis
 url: abc.com
 timeout: 30000ms
```

akelad:
Thanks for raising this, it's definitely a good idea. If it's something you'd be interested in implementing, we'd happily accept a pull request

souvikg10:
I will try :D

PR #1048, opened 2018-09-22
Tracker_store added in endpoints.yml

souvikg10:
**Proposed changes**:
- A way to externalise the tracker_store from default(InMemory) to either Redis or MongoDB using the endpoints.yml
You can add the config for the Redis or MongoDB in the endpoints.yml file, mentioning the store_type either 'redis' or 'mongod' 

This will help not to add new code in order to externalise the tracker store

Issue #1035 

- [x] updated the changelog


akelad:
Thanks for submitting the PR, we'll review it as soon as possible 🙂

MetcalfeTom:
Love the functionality!  Will require a little more implementation - [please check out the `store.store_type` error raised by Travis here](https://travis-ci.com/RasaHQ/rasa_core/jobs/147333224#L1581)

souvikg10:
I wonder why this is throwing an error though for InMemoryTracker, I didn't change anything for that

souvikg10:
Nevermind, i see the issue 🗡 , I will update it

rasa_core/utils.py
```diff
@@ -595,27 +595,34 @@ def read_endpoints(cls, endpoint_file):
...
         self.nlg = nlg
+        self.tracker_store = tracker_store
 
 
 class EndpointConfig(object):
     """Configuration for an external HTTP endpoint."""
 
     def __init__(self, url, params=None, headers=None, basic_auth=None,
-                 token=None, token_name="token"):
+                 token=None, token_name="token",store_type=None,db=None,user=None,password=None,timeout=None):
```
tmbo: I am not sure we should set all of them as separate arguments. Now that we start to use this for more purposes, I think it makes sense to just take all the `**kwargs` and store them on the objects property dict:
```
for k, v in kwargs.items():
  setattr(self, k, v)
```

souvikg10: yeah makes sense, i had some thoughts on that as well. I will try to do that next week 👍

rasa_core/agent.py
```diff
@@ -648,8 +648,7 @@ def create_tracker_store(store, domain):
             store.domain = domain
             return store
         else:
-            return InMemoryTrackerStore(domain)
-
+            return InMemoryTrackerStore(domain)
```
wochinge: please add a blank line after the method

rasa_core/tracker_store.py
```diff
@@ -30,7 +30,14 @@ def __init__(self, domain, event_broker=None):
         # type: (Optional[Domain], Optional[EventChannel]) -> None
         self.domain = domain
         self.event_broker = event_broker
-
```
wochinge: please re-add an empty line here

rasa_core/tracker_store.py
```diff
@@ -124,6 +131,7 @@ def __init__(self, domain, host='localhost',
                  port=6379, db=0, password=None, event_broker=None,
                  record_exp=None):
 
+
```
wochinge: please remove the empty line

rasa_core/train.py
```diff
@@ -216,7 +217,8 @@ def train_dialogue_model(domain_file, stories_file, output_path,
     _endpoints = AvailableEndpoints.read_endpoints(cmdline_args.endpoints)
     _interpreter = NaturalLanguageInterpreter.create(cmdline_args.nlu,
                                                      _endpoints.nlu)
-
+    domain = TemplateDomain.load(cmdline_args.domain)                                                
+    _tracker_store = TrackerStore(domain).find_tracker_store(_endpoints.tracker_store)
```
wochinge: please insert an empty line here for better readability

wochinge:
If you fix the error in `rasa_core.utils.py` and a few formatting issues we should be good to go with your pull request! 🚀 I didn't comment on every formatting issue, Travis should give you feedback on that 😄
Thanks for your effort!

rasa_core/utils.py
```diff
@@ -678,7 +681,12 @@ def from_dict(cls, data):
                 data.get("headers"),
                 data.get("basic_auth"),
                 data.get("token"),
-                data.get("token_name"))
```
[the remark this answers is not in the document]
wochinge: alternatively:

```
EndpointConfig(
                data.get("url"),
                data.get("params"),
                data.get("headers"),
                data.get("basic_auth"),
                data.get("token"),
                data.get("token_name"),
                **data)
```

rasa_core/tracker_store.py
```diff
@@ -30,7 +30,14 @@ def __init__(self, domain, event_broker=None):
         # type: (Optional[Domain], Optional[EventChannel]) -> None
         self.domain = domain
         self.event_broker = event_broker
-
+    def find_tracker_store(self, store=None):
+        if store is None:
+            return InMemoryTrackerStore(self.domain)
+        elif store.store_type == 'redis':
+            return RedisTrackerStore(domain=self.domain,host=store.url,db=store.db,password=store.password,record_exp=store.timeout)
```
wochinge: It would be great if the user could also define a certain port through the configuration file.
Hence, we should probably change this line to:
```python
return RedisTrackerStore(domain=self.domain, 
                         host=store.url, 
                         port=store.port,
                         db=store.db, 
                         password=store.password, 
                         record_exp=store.timeout)

```

rasa_core/utils.py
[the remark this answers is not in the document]
souvikg10: that makes sense. I will do the fixes this weekend 👍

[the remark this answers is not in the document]
souvikg10: I updated the issues, however i haven't found any formatting error in Travis, not sure how to check that though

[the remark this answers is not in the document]
wochinge: Thanks, I will have another look on them later. Regarding travis: I think we have to let it built first. This takes usually about 20 minutes after you have pushed your changes.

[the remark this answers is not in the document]
wochinge: A solution to this could be:

```python
return EndpointConfig(**data)
```

and a constructor like this:

```python
class EndpointConfig(object):
    """Configuration for an external HTTP endpoint."""

    def __init__(self, url, params=None, headers=None, basic_auth=None,
                 token=None, token_name="token",**kwargs):
        self.url = url
        self.params = params if params else {}
        self.headers = headers if headers else {}
        self.basic_auth = basic_auth
        self.token = token
        self.token_name = token_name

        [setattr(self, k, v) for k, v in kwargs.items() if not hasattr(self, k)]
```

I used the `if not hasattr(self, k)` to not overwrite the dictionaries of `self.params` and `self.headers` with None.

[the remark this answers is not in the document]
wochinge: Regarding the formatting issues: You can also check that locally, e.g. by using a linter which knows pep8 or IDEs such as PyCharm or Visual Studio Code (with the Python plugin).

wochinge:
Hi @souvikg10,
sorry for the late feedback but Github was unusable due to their problems yesterday.
Your solution looks pretty good to me now! Thanks for that 👍 

A few more things:
- We discovered why you could not see the pep8 issues in Travis. To do see them, please rebase your branch on the master and push again. Travis should then properly run the pep8 validations.
- Add a few tests (especially for `find_tracker_store` and for your changes in `utils.py` (e.g. that the values are correctly set, `params` and `headers` are empty dictionaries instead of `None`).

Thanks 🙏

akelad:
@souvikg10 i just hit the `update branch` button for you 😄 that should show the pep8 errors

souvikg10:
I added some tests for find_tracker_store. I fixed pep8 issues discovered by travis and my vscode, however, I noticed there are a lot of error linked to 
```E501:line too long (80 > 79 characters)```
I wasn't sure what to do about them, since this rules could differ

**Comment by wochinge** [comment]
Please wrap the lines so that no line is longer than 80 characters otherwise Travis will fail.
I will checkout your other changes tomorrow but it looks like we are ready to merge then 🚀👍

**Comment by wochinge** [comment]
I just saw in Travis that `test_endpoint_config` is failing because you specified the `url` param multiple times.

File: rasa_core/train.py [hunk 8]
```diff
@@ -139,7 +140,7 @@ def create_argument_parser():
             '--fallback_action_name',
             type=str,
             default=DEFAULT_FALLBACK_ACTION,
-            help="When a fallback is triggered (e.g. because the ML prediction "
+            help="When a fallback is triggered (e.g. because the ML prediction"
```
wochinge: please add a space after `prediction` otherwise it's `because of the ML predictionis of low confidence`

rasa_core/utils.py
```diff
@@ -486,7 +486,7 @@ def extract_args(kwargs,  # type: Dict[Text, Any]
 
 
 def arguments_of(func):
-    """Return the parameters of the function `func` as a list of their names."""
+    """Return the parameters of the function `func` as a list of their names"""
```
wochinge: please re-add the point at the end. If it then success 80 chars, wrap the line.

rasa_core/train.py
souvikg10: I think it was crossing the 80 character limit

wochinge: then wrap it differently, e.g. 
```python
help="When a fallback is triggered (e.g. because the ML "
                 "prediction is of low confidence) this is the name of the "
                 "action that will get triggered instead."
```

There is also a typo in line 144 (tje -> the). I know it was not caused by you, but would be cool if you just fix it together with the missing space.

souvikg10: okay no problem.

wochinge:
Thanks a lot @souvikg10 . This looks really great now 🚀 👍 
One thing which I discovered when testing it out: It might be a nice convenience to provide default values for parameters like `port`, `password` (especially for redis since we cannot set it to `None` otherwise`), `record_exp`. Hence, the user would not have to provide or think about all these values when they don't differ from the regular setup.

@tmbo @MetcalfeTom If you also want to review this, then now would be a good time.

souvikg10:
Where should the defaults be set? find_tracker_store?

souvikg10:
I checked the class RedisTrackerStore and i see in the __init__ , defaults are set for port as 6379 and db as 0 and password as None , so i suppose if you don't pass them via the endpoints, these will be set during init class

wochinge:
@souvikg10 The problem is that in `find_tracker_store` attributes are accessed which might not be present. E.g. if you don't provide a port in the endpoints for Redis line 40 would throw an exception as `store.port` accesses an attribute which is not present. Since the actual TrackerStores, like `RedisTrackerStore` are providing default values, it would be enough to pass `None` as parameter if the attribute is not present.  This could be done using `getattr` but there might be smarter ways to do it.

souvikg10:
I see, it could be useful to set defaults there. makes sense :+1: i will add them

wochinge:
Thanks! You are the best 👍

wochinge:
And please add a note to the changelog! I have already written the documentation so no need for that.

souvikg10:
I created a new method that checks if attributes are present from Endpoint, if not sets to None so tracker_store gets it as default. except for URL, which is mandatory. I did this in tracker_store.py making sure it is agnostic to any tracker store. However if new tracker stores are added which needs more params, then defaults needs to be set. Not sure how to avoid that

rasa_core/tracker_store.py
```diff
@@ -31,6 +31,38 @@ def __init__(self, domain, event_broker=None):
...
+            store.db = None
+        if not hasattr(store, "username"):
+            store.port = None
+        if not hasattr(store, "password"):
+            store.password = None
+        if not hasattr(store, "timeout"):
+            store.timeout = None
+        return store
+
+    def find_tracker_store(self, store=None):
```
wochinge: I just realised that this should be class method. We are just using the `domain` attribute which could just be a parameter of `find_tracker_store`.

rasa_core/utils.py
```diff
@@ -601,27 +602,37 @@ def read_endpoints(cls, endpoint_file):
...
-                 token=None, token_name="token"):
+                 token=None, token_name="token", **kwargs):
         self.url = url
         self.params = params if params else {}
         self.headers = headers if headers else {}
         self.basic_auth = basic_auth
         self.token = token
         self.token_name = token_name
+        [setattr(
+            self, k, v) for k, v in kwargs.items() if not hasattr(self, k)]
```
wochinge: If we replace these lines with 
```
self.store_type = kwargs.pop('store_type', None)
self.kwargs = kwargs
```

then we can simplify this whole default parameter thing -> see `find_tracker_store` for the follow up

rasa_core/tracker_store.py
```diff
@@ -31,6 +31,38 @@ def __init__(self, domain, event_broker=None):
...
+        if store is None:
```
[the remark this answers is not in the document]
souvikg10: Indeed, then timeout needs to be replaced with record_exp in the Endpoints file. Also to update the documentation as well.
I will take care of that

rasa_core/utils.py
souvikg10: okay, i see the other review. It will automatically be passed to the Redis or Mongo Tracker

wochinge: If we pass the extra tracker store parameters as dictionary and then pass the dictionary as key word arguments (`**kwargs`) to the actual constructor of a tracker store, then the actual tracker store should provide the default values. Was that understandable?

wochinge: Ah, you were faster 👍

wochinge:
@souvikg10 The suggested approach is open for discussion. It's just what I came up with when I read your previous comment.

souvikg10:
your approach looks cleaner. I was wondering this yesterday about what happens if there is a new custom tracker_store which is linked to your enhancement request regarding adding any tracker store in the future

wochinge:
Yes, that's exactly the same I thought. If got this great possibility to define everything in the `endpoints.yml` then it would be quite cumbersome to have to adjust the `check_attributes` all the time and set values to `None` we might not even need.

rasa_core/tracker_store.py
```diff
@@ -31,6 +31,20 @@ def __init__(self, domain, event_broker=None):
         self.domain = domain
         self.event_broker = event_broker
 
+    @classmethod
```
wochinge: I think `staticmethod` should also be fine since we don't require access to any attributes / state of the class.

souvikg10: okay, i can change that

rasa_core/run.py
```diff
@@ -245,10 +246,13 @@ def load_agent(core_model, interpreter, endpoints,
     _endpoints = AvailableEndpoints.read_endpoints(cmdline_args.endpoints)
     _interpreter = NaturalLanguageInterpreter.create(cmdline_args.nlu,
                                                      _endpoints.nlu)
+    domain = TemplateDomain.load(os.path.join(cmdline_args.core, "domain.yml"))
+    _tracker_store = TrackerStore(domain).find_tracker_store(
```
wochinge: Since we now are using the `@staticmethod`
```python
TrackerStore.find_tracker_store(domain, _endpoints.tracker_store)
```

rasa_core/train.py
```diff
@@ -232,6 +233,9 @@ def train_dialogue_model(domain_file, stories_file, output_path,
     _interpreter = NaturalLanguageInterpreter.create(cmdline_args.nlu,
                                                      _endpoints.nlu)
 
+    domain = TemplateDomain.load(cmdline_args.domain)
+    _tracker_store = TrackerStore(domain).find_tracker_store(
```
wochinge:
```
TrackerStore.find_tracker_store(domain, _endpoints.tracker_store)
```

souvikg10: jeez i missed it :P sorry about that

wochinge: Absolutely no problem. Thanks for being so responsive and contributing!

rasa_core/run.py
```diff
@@ -245,10 +246,13 @@ def load_agent(core_model, interpreter, endpoints,
     _endpoints = AvailableEndpoints.read_endpoints(cmdline_args.endpoints)
     _interpreter = NaturalLanguageInterpreter.create(cmdline_args.nlu,
                                                      _endpoints.nlu)
+    domain = TemplateDomain.load(os.path.join(cmdline_args.core, "domain.yml"))
```
tmbo: this file might not exist: the agent might be loaded from a remote server and this directory might be empty 

It is fine to set the `TrackerStore`'s domain to `None` in line L250 and not load the domain here (the actual domain will be updated once it is loaded) - though we really need to fix this...

rasa_core/train.py
```diff
@@ -232,6 +233,9 @@ def train_dialogue_model(domain_file, stories_file, output_path,
     _interpreter = NaturalLanguageInterpreter.create(cmdline_args.nlu,
                                                      _endpoints.nlu)
 
+    domain = TemplateDomain.load(cmdline_args.domain)
+    _tracker_store = TrackerStore.find_tracker_store(domain,
```
tmbo: same here, better to set the `domain` to `None` and not load it

tmbo:
I've added a comment about the domain loading, wich we need to remove. Otherwise it will fail for agents stored on remote servers. 

But other than that it is a :rocket: from my side. Thanks a lot for putting so much effort into this @souvikg10

rasa_core/train.py
souvikg10: I removed the domain dependency and changed it to None, ran the tests again and it was fine.

tmbo: thanks a lot :+1:

rasa_core/tracker_store.py
```diff
@@ -31,6 +31,20 @@ def __init__(self, domain, event_broker=None):
         self.domain = domain
         self.event_broker = event_broker
 
+    @staticmethod
+    def find_tracker_store(domain, store=None):
+        if store is None or store.store_type is None:
```
MetcalfeTom: `if None in {store, store.store_type}`

rasa_core/tracker_store.py
```diff
@@ -31,6 +31,20 @@ def __init__(self, domain, event_broker=None):
...
+            return InMemoryTrackerStore(domain)
+        else:
```
MetcalfeTom: Delete the line below and replace this with `elif`

rasa_core/tracker_store.py
tmbo: I think this one is controversial, I'd argue the `or` way is easier to understand

MetcalfeTom: OK then, leave it out

rasa_core/tracker_store.py
souvikg10: Updated it

wochinge:
@MetcalfeTom then we are ready to merge?

MetcalfeTom:
Congratulations on another successful merge @souvikg10

PR #1108, opened 2018-10-01
Add training date and time to policy metadata when persisting  Fixes #1104

### LINKED ISSUE #1104: persist trained_at in metadata
rasa core should persist trained_at in metadata so that platform can show the true date not the upload date

MetcalfeTom:
- fixes #1104

**Proposed changes**:
- Added date to _persist_metadata() function of base policy ensemble

- [x] updated the changelog

### DIFF
--- a/rasa_core/policies/ensemble.py
+++ b/rasa_core/policies/ensemble.py
@@ -8,7 +8,7 @@ import json
 import logging
 import os
 import sys
-import time
+import datetime
 from collections import defaultdict
 
 import numpy as np
@@ -40,7 +40,6 @@ class PolicyEnsemble(object):
         # type: (List[Policy], Optional[Dict]) -> None
         self.policies = policies
         self.training_trackers = None
-        self.date_trained = None
 
         if action_fingerprints:
             self.action_fingerprints = action_fingerprints
@@ -127,7 +126,7 @@ class PolicyEnsemble(object):
             "max_histories": self._max_histories(),
             "ensemble_name": self.__module__ + "." + self.__class__.__name__,
             "policy_names": policy_names,
-            "trained_at": self.date_trained
+            "trained_at": datetime.datetime.now().strftime('%Y%m%d-%H%M%S')
         }
 
         utils.dump_obj_as_json_to_file(domain_spec_path, metadata)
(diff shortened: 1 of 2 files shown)


rasa_core/policies/ensemble.py
```diff
@@ -123,7 +126,8 @@ def _persist_metadata(self, path, dump_flattened_stories=False):
             "python": ".".join([str(s) for s in sys.version_info[:3]]),
             "max_histories": self._max_histories(),
             "ensemble_name": self.__module__ + "." + self.__class__.__name__,
-            "policy_names": policy_names
+            "policy_names": policy_names,
+            "trained_at": self.date_trained
```
ricwo: can we just use the timestamp at persistence time as this happens right after actual training? in some sense the `persist()` method completes the training. in that case we wouldn't need the `self.date_trained` attribute

rasa_core/policies/ensemble.py
```diff
@@ -65,6 +67,7 @@ def train(self, training_trackers, domain, **kwargs):
             for policy in self.policies:
                 policy.train(training_trackers, domain, **kwargs)
             self.training_trackers = training_trackers
+            self.date_trained = time.strftime("%Y-%m-%d %H:%M:%S", time.gmtime())
```
ricwo: I'm not sure about `gmtime()` as it might be confusing. can we not use the machine time? I'd suggest we also keep the timestamp format the same as in NLU metadata
https://github.com/RasaHQ/rasa_nlu/blob/master/rasa_nlu/model.py#L116

MetcalfeTom: Excellent idea, without your prior knowledge I would not have picked up on this use before

rasa_core/policies/ensemble.py
MetcalfeTom: I'm not sure. I agree that persist completes the training but having it inside the training function ensures that the time is logged as it is described (in case the script in question doesn't follow `.train()` with `.persist()` immediately after).

ricwo: You're right, but that's very much a special case. either way works, really

PR #1127, opened 2018-10-04
Interactive learning with graph

### LINKED ISSUE #1133: Interactive learning: restart should behave the same way as "start fresh"
right now the previous conversation gets wiped, but when doing `/restart` this should still keep the previous conversations and export them all at the end (same as the "start fresh" menu option)

tmbo:
**Proposed changes**:
- shows the plotted graph in the browser during interactive learning
- `RegexInterpreter` ensures that the user message has the `INTENT_PREFIX` (e.g. `/`)
- implemented partial visualization "around" a tracker (so only the closer neighborhood of the story graph is visualized)
- removed `/stop` intent - user should use `Crtl-c` instead 
- treat user input `/restart` as if the user has selected "Restart" from the exit menu. fixes #1133
- pin the version of `prompt_toolkit `

**TODO**
- [ ] add a couple tests for the visualization functions

**Test instructions**
To test this, run 
```
python rasa_core/train.py -s examples/moodbot/data/stories.md \
  -o  examples/moodbot/models/dialogue --domain examples/moodbot/domain.yml \
  --interactive --epochs 2
```
after checking out the branch.

ezgif com-video-to-gif

- [x] updated the changelog


wochinge:
Should the red circled part in this case not also be blue box with a question mark? image

wochinge:
The question is maybe also how we handle None intents: image

tmbo:
The first issue is fixed :+1:

@wochinge  About the second one: what do you mean with a `None Action` ?

wochinge:
Sorry, it the naming is wrong. I mean a `None` intent, e.g. if the user just sends an empty message.

Another, possible improvement: When the bot is listening for user input, the arrow to the leftmost possible intent is *bold*. This might confuse the user as the user might think that this then is the default or most likely action

tmbo:
I added some handling for the `None` case. will be treated as if the user had pressed Ctrl-c, because a `None` message doesn't really make sense.

concerning the second part: do you mean the leftmost arrow should be gray as well?

bildschirmfoto 2018-10-16 um 17 12 44

wochinge:
Yes, I think it should be gray until the user actually made another input.

tmbo:
I like the suggestion and took a look at it - to complicated for the moment 😅 . Let's keep it in mind and see what other users say - anything else you noticed?

wochinge:
Yes, sometimes I am running in loops where the bot keeps on uttering stuff and is not listening to user input anymore. This happens when I execute `/greet` several times. But I do not know whether this is related to this pull request or simply a problem because of missing training data.
image

tmbo:
Good point, but thats more an issue of the trained model (the model doesn't really get trained, in the above example command it only gets trained for 2 epochs, which is way to few for this small dataset)

Issue #1135, opened 2018-10-05
error in docs for RestInput


amn41:
**Rasa Core version**: 0.11.9

**Issue**:

Docs state that the payload for the `RestInput` channel should be:

```
{
  "sender": <>,
  "text": <>
}
```

but the correct payload is:

```
{
  "sender": <>,
  "message": <>
}
```
**Content of domain file** (if used & relevant):
```yaml

```

Zhipeng-Chang:
Hi! I would like to work on this issue, I just wonder which doc contains the wrong payload?

Zhipeng-Chang:
[https://github.com/RasaHQ/rasa_core/pull/1139](url)

Zhipeng-Chang:
https://github.com/RasaHQ/rasa_core/pull/1139

ankitameht:
Hi I can work on this issue, if @Zhipeng-Chang is not already on it! :)

ankitameht:
Great!

PR #1121, opened 2018-10-04
look for username attr rather than user 

amn41:
**Proposed changes**:
- fixes #1120 

**Status (please check what you already did)**:
- [x] made PR ready for code review
- [ ] added some tests for the functionality
- [ ] updated the documentation
- [ ] updated the changelog

PR #1139, opened 2018-10-05
fix "error in docs for RestInput"

Zhipeng-Chang:
**Proposed changes**:
- ...

- [x] updated the changelog


Zhipeng-Chang:
Since no new tests need to be added, checked " added some tests for the functionality"

Zhipeng-Chang:
link to the issue: [](https://github.com/RasaHQ/rasa_core/issues/1135)

Zhipeng-Chang:
fix another bug "wrongly extracting username from JWT #1140" https://github.com/RasaHQ/rasa_core/issues/1140

akelad:
@Zhipeng-Chang Thanks for submitting this PR! looks like your PR is based on a protected branch? So I can't update it or anything, can you please change that? 
Also, an initial comment: could you remove the changes from the Changelog? Or rather summarise them in one line please. And also reapply the username change please, i'll figure out what's going wrong with the build

Zhipeng-Chang:
Hi @akelad Done!

akelad:
@Zhipeng-Chang ok it seems the reason the build is failing is because the test `test_get_tracker_with_jwt` is now using the wrong fake JWT token. In the comment there you can see that it's still in the old format `{"user": {"user": "testadmin", "role": "admin"}}`, which should be changed to `{"user": {"username": "testadmin", "role": "admin"}}`. You can also replace the token in there with `eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyIjp7InVzZXJuYW1lIjoidGVzdGFkbWluIiwicm9sZSI6ImFkbWluIn19.NAQr0kbtSrY7d28XTqRzawq2uQRre7IWTuIDrCn5AIw`

rasa_core/server.py
```diff
@@ -385,7 +385,7 @@ def predict(sender_id):
     @ensure_loaded_agent(agent)
     def log_message(sender_id):
         request_params = request.get_json(force=True)
-        message = request_params.get("text")
+        message = request_params.get("message")
```
tmbo: I don't think we can just change this.

Zhipeng-Chang: Hi @tmbo ! Thanks for reviewing this! Could you please provide more resons for this?

tmbo: well, it is a breaking api change as anyone who is using this api needs to change their payloads. I see the benefit of the change, making things more consistent - but it would be better to make this backwards compatible.

Zhipeng-Chang: Hi! I just added a "try except" handler for this!

Zhipeng-Chang:
Hi @akelad! 
I changed the "testadmin" token to the one as you provided, but the build server is still failing. I looked into the build server log and found that the the "testuser" token should be updated as well, eg. { "user": { "username": "testuser", "role": "user" } }. I created the token by myself, but I got the return code of 422 from the build server, I copied the error message from Travis-ci "       
# {"user": {"username": "testuser", "role": "user"}}
        jwt_header = {
            "Authorization": "Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9."
                             "eyJ1c2VyIjp7InVzZXJuYW1lIjoidGVzdHVzZXIiLCJ"
                             "yb2xlIjoidXNlciJ9fQ.EPJ8zMVdQPqnNvM0zkTO"
                             "NS-M8ptfOv0kbqMZqcauYSI"
        }
        response = secured_app.get("/conversations/testadmin/tracker",
                                   headers=jwt_header)
>       assert response.status_code == 403
E       assert 422 == 403
E        +  where 422 = <Response streamed [422 UNPROCESSABLE ENTITY]>.status_code"

I think maybe the token created by me is not encoded properly? so something else ... would you mind to create a token for testuser and send back to me?

Thank you.

Best,

akelad:
Hey @Zhipeng-Chang sorry about all the back and forth, but looks like we actually have another PR open for the `username` thing: #1121 
Would you mind removing those changes and then we can take a look at the other ones you made?

tests/test_server.py
```diff
@@ -267,7 +266,6 @@ def test_get_tracker_with_jwt(secured_app):
                                headers=jwt_header)
     assert response.status_code == 200
 
-
```
akelad: please put these newlines back in, the pep8 check fails otherwise

rasa_core/server.py
```diff
@@ -385,7 +385,11 @@ def predict(sender_id):
     @ensure_loaded_agent(agent)
     def log_message(sender_id):
         request_params = request.get_json(force=True)
-        message = request_params.get("text")
+        try:
```
akelad: @tmbo are you happy with this way of dealing with it? we should probably except the specific error, like `KeyError` or whatever

tmbo: this won't work (there is not going to be a key error, as `.get` never throws one), should be 

```
try:
  message = request_params["message"]
except KeyError:
  message = request_params.get("text")
```

Zhipeng-Chang: Hi @tmbo, Yes! you are right! As I just tried with json.get function. If the value does not exist, python will return "None" instead of throwing an error. Thanks for pointing this out!

Zhipeng-Chang:
Thanks @akelad !

PR #1153, opened 2018-10-09
fix putting `action_restart` in the stories break training #906

Ghostvv:
**Proposed changes**:
- fix #906

- [ ] updated the changelog

### DIFF
--- a/rasa_core/training/generator.py
+++ b/rasa_core/training/generator.py
@@ -232,7 +232,8 @@ class TrainingDataGenerator(object):
         # checkpoints that seem to be reachable. This is a heuristic,
         # if we did not reach any new checkpoints in an iteration, we
         # assume we have reached all and stop.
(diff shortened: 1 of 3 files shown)


rasa_core/training/generator.py
```diff
@@ -469,16 +474,18 @@ def _process_step(
                     new_sender = step.block_name
                 trackers.append(tracker.copy(new_sender))
 
-        new_trackers = []
+        end_trackers = []
```
tmbo: not sure about this naming, because the conversation is not finished at this point, is it?

Ghostvv: yes, but in `deduplication` the conversation didn't finish as well, but we call it `end_trackers` because we do not add any checkpointed story blocks to them. But if you have any suggestion, I'm open to change the name

tmbo: all good

PR #1155, opened 2018-10-09
Update class' terminology from Rasa SDK

ruiconti:
**Proposed changes**:
- Fixed CustomAction's `run()` documentation. Now it receives a Tracker not a DialogueStateTracker

- [ ] updated the changelog

### DIFF
--- a/docs/customactions.rst
+++ b/docs/customactions.rst
@@ -104,7 +104,7 @@ which might look like this:
          return "action_check_restaurants"
 
       def run(self, dispatcher, tracker, domain):
-         # type: (Dispatcher, DialogueStateTracker, Domain) -> List[Event]
+         # type: (Dispatcher, Tracker, Domain) -> List[Event]
 
          cuisine = tracker.get_slot('cuisine')
          q = "select * from restaurants where cuisine='{0}' limit 1".format(cuisine)


akelad:
thanks for submitting this PR! but actually `DialogueStateTracker` is the correct one: https://github.com/RasaHQ/rasa_core/blob/master/rasa_core/trackers.py#L49
So i'll close this PR

ruiconti:
@akelad Isn't it rasa_core_sdk's `Action` object used on the dialogue transaction? From my experience with the most updated version, I get `Tracker` instead of `DialogueStateTracker` object transactioned within a CustomAction:
https://github.com/RasaHQ/rasa_core_sdk/blob/master/rasa_core_sdk/__init__.py#L153

Ghostvv:
@ruiconti thanks for noticing it, you are right it should be `Tracker`. Could you please update other types as well, as in: https://github.com/RasaHQ/rasa_core_sdk/blob/be14ce292e084c6914376b70eee1f7858153a11c/rasa_core_sdk/__init__.py#L153

docs/customactions.rst
```diff
@@ -104,7 +104,7 @@ which might look like this:
          return "action_check_restaurants"
 
       def run(self, dispatcher, tracker, domain):
-         # type: (Dispatcher, DialogueStateTracker, Domain) -> List[Event]
+         # type: (Dispatcher, Tracker, Domain) -> List[Event]
```
Ghostvv: Could you please fix other types as well: `# type: (CollectingDispatcher, Tracker, Dict[Text, Any]) -> List[dict]`

ruiconti:
In order to sphinx successfully reference `rasa_core_sdk.Action.run()`, its documentation also needed update.
I made another PR to fix that: https://github.com/RasaHQ/rasa_core_sdk/pull/15

akelad:
ok actually it seems you can't reference a rasa_core_sdk class from the docs. @tmbo any workaround?

tmbo:
yes, the sdk needs to be installed in the docs build stage (it currently is only installed in the test stage when installing the requirements)

akelad:
@ruiconti one more thing to fix then, sorry 😄

PR #1158, opened 2018-10-11
Fix serve_forever=True for all tested channels

arcontechnologies:
the server_forever was set to True for keeping the server running.

**Proposed changes**:
- ...

- [ ] updated the changelog


arcontechnologies:
updated (only comment) to avoid travis builds fail

PR #1175, opened 2018-10-16
Update Interactive learning documentation

samarth12:
The documentation doesn't mention how to use a trained nlu model for intent classification with interactive learning. Although it is an intuitive command, people might not be able to figure it out so easily and just stay confused about why the intents from their model are not showing up.

**Proposed changes**:
- ...

- [ ] updated the changelog

### DIFF
--- a/docs/interactive_learning.rst
+++ b/docs/interactive_learning.rst
@@ -29,8 +29,7 @@ Run the following to start interactive learning:
    python -m rasa_core.train \
      --interactive -o models/dialogue \
      -d domain.yml -s stories.md \
-     --endpoints endpoints.yml \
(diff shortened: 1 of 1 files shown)


docs/interactive_learning.rst
```diff
@@ -29,7 +29,10 @@ Run the following to start interactive learning:
    python -m rasa_core.train \
      --interactive -o models/dialogue \
      -d domain.yml -s stories.md \
-     --endpoints endpoints.yml
+     --endpoints endpoints.yml \
+     --nlu models/current/nlu (OPTIONAL)
```
akelad: can you remove the flag here? I think it's enough to have the comment below

Issue #1130, opened 2018-10-04
Tracker issues


amsolu:
**Rasa Core version**:0.11.5

**Python version**: 3.6

**Operating system** (windows, osx, ...): Linux Ubuntu

**Issue**: Using the tracker to save conversations.
I was trying to get the conversations from the different channels I did connect to my Rasa Stack
Using http://localhost:5002/conversations I did get the list of active conversations however I am not able to get the information from Telegram conversations, I found out Telegram is the single one represented as a number instead of a text, see below:

(bot36) ale@ubuntu:~/Projects/weather_bot$ curl http://localhost:5002/conversations | python -mjson.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    72  100    72    0     0  10285      0 --:--:-- --:--:-- --:--:-- 10285
[
    "+130XXXX1311",        <-  Twilio
    626931294,               <- Telegram
    "2270705826305018" <- facebook
]

Any time I am running twilio and facebook tracker with its sender_id works but when trying the same with Telegram, I am getting this:

(bot36) ale@ubuntu:~/Projects/weather_bot$ curl http://localhost:5002/conversations/626931294/tracker | python -mjson.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   270  100   270    0     0   2142      0 --:--:-- --:--:-- --:--:--  2142
{
    "events": [
        {
            "event": "action",
            "name": "action_listen",
            "timestamp": 1538625454.9052396
        }
    ],
    "followup_action": null,
    "latest_event_time": 1538625454.9052396,
    "latest_message": {
        "entities": [],
        "intent": {},
        "text": null
    },
    "paused": false,
    "sender_id": "626931294",
    "slots": {
        "location": null
    }
}
Rasa is adding the quotes and I am assuming this is the reason I am getting all empty.
If I am running conversations again, a new sender_id appear which looks like the same Telegram id including quotes.

I don't know where the issue is, but I have no more ideas :(
(bot36) ale@ubuntu:~/Projects/weather_bot$ curl http://localhost:5002/conversations | python -mjson.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
[...]

PR #1176, opened 2018-10-16
#1130 handle telegram user ids internally as `str`

wochinge:
#1130 handle telegram user ids internally as `str` as the ids we receive from other services

**Motivation**:
The telegram library parses the user id as `int`. However, rasa_core implicitly expects a `str`.
When a user queries the conversations for a specific user id, the input is processed as `str`. Rasa returns an empty conversation list, as the tracker only contains entries for the user id as `int` but not as `str`.

**Proposed changes**:
By casting the user id to a `str`, we handle the Telegram user id equally to the ids of other services.
We do not have to parse back the user id to a `int` when sending message back to Telegram, as the library accepts user ids as `str` and `int`.

- [ ] updated the changelog


wochinge:
Sorry, I totally overlooked that. I will redo that then.

wochinge:
@akelad I updated my commit based on the conversation in Slack.

rasa_core/tracker_store.py
```diff
@@ -13,6 +13,8 @@
 import six.moves.cPickle as pickler
 from typing import Text, Optional, List
 
+from pymongo import ReturnDocument
```
akelad: please move this import to the retrieve function, since it's only actually needed for the MongoTrackerStore

rasa_core/tracker_store.py
```diff
@@ -192,6 +194,13 @@ def save(self, tracker, timeout=None):
 
     def retrieve(self, sender_id):
         stored = self.conversations.find_one({"sender_id": sender_id})
+
+        # look for conversations which have used an `int` sender_id in the past and update them.
+        if stored is None and sender_id.isdigit():
+            stored = self.conversations.find_one_and_update({"sender_id": int(sender_id)},
```
akelad: Just checking, this returns None if no such document is found right?

wochinge: I actually tried it out since the documentation of pymongo is about unclear about that.
image

wochinge:
@akelad Updated according to your suggestions.

PR #1186, opened 2018-10-17
New actions interactive learning

### LINKED ISSUE #1184: support new actions during interactive learning
It would be great to support creating new actions during interactive learning. The user would come up with the new action on the spot, but would have to back later and ensure there is a template defined (for utter actions) or there is a custom action registered at the appropriate endpoint.

amn41:
**Proposed changes**:
WIP
- fixes #1184 

- [ ] updated the changelog


amn41:
1. current implementation asks user for a new action name, and if provided just logs an `actionexecuted` event on the tracker, as opposed to attempting to execute a (possibly non-existent) action. However this has the downside that if your action would return events when executed, the story you create in IL is not correct. Alternative would be to first attempt execution and then fall back to simply logging the action, while throwing a user warning that their story may be wrong. 

2. since we allow the user to define new intents/ entities/ actions on the fly we should probably also support dumping an updated `domain.yml` at the end of training.

@tmbo @MetcalfeTom @akelad thoughts?

MetcalfeTom:
1. Yes, seems better than the current version.  I would suggest throwing this user warning up with the action request prompt.

2. Good idea.  Do we need to do this at the end of the session though?  It would be quicker to append a new child element to the YAML file at the end of `_request_intent_from_user`

tmbo:
Yes I think we should add it to the domain - should we ask the user for the events of that action? might be to cumbersome

amn41:
tests fail, but it's the same 3 tests as on the target branch (`forms_v2`). 

TODO: make dumped domain file pretty

amn41:
@tmbo as you suggested during community event, it would be cool to also support adding utterances for new actions directly. (though then we go down the rabbit hole of buttons, images ...)

amn41:
closing in favour of https://github.com/RasaHQ/rasa_core/pull/1224

PR #1189, opened 2018-10-18
Core only dockerfile

wochinge:
**Motivation**:
Issue: https://github.com/RasaHQ/rasa_core/issues/845

**Proposed changes**:
- remove nlu from rasa core Dockerfile
- add docker compose file which straps together core, nlu, core sdk and duckling

- [x] updated the changelog

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -36,5 +36,10 @@ services:
     volumes:
     - "./rasa-app-data/projects/:/app/projects"
     - "./rasa-app-data/logs:/app/logs"
+  duckling:
+    image: rasa/duckling:latest
+    networks: ['rasa-network']
+    ports:
+    - "8000:8000"
 
 networks: {rasa-network: {}}
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,20 +1,20 @@
 version: '3.0'
 
 services:
-  rasacore:
+  rasa_core:
       image: rasa/rasa_core:latest
       networks: ['rasa-network']
       ports:
       - "5005:5005"
       volumes:
-      - "./rasa-app-data/core-model:/app/model"
+      - "./rasa-app-data/models/current/dialogue:/app/model"
       - "./rasa-app-data/config:/app/config"
       - "./rasa-app-data/project:/app/project"
       command: 
         - start
         - -d
         - ./model
-        - -c 
+        - -c
         - rest
         - -u
         - current/nlu
@@ -27,13 +27,13 @@ services:
       - "5055:5055"
       volumes:
       - "./rasa-app-data/actions:/app/actions"
-  rasanlu:
+  rasa_nlu:
     image: rasa/rasa_nlu:latest-full
     networks: ['rasa-network']
     ports:
     - "5000:5000"
     volumes:
-    - "./rasa-app-data/nlu-models/:/app/projects"
+    - "./rasa-app-data/models/:/app/projects"
     - "./rasa-app-data/logs:/app/logs"
   duckling:
     image: rasa/duckling:latest
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,0 +1,40 @@
+version: '3.0'
+
+services:
+  rasacore:
+      image: rasa/rasa_core:latest
+      networks: ['rasa-network']
+      ports:
+      - "5005:5005"
+      volumes:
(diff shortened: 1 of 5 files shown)


wochinge:
Ok, I add the documentation.
Should we add an extra volume in the `Dockerfile` for the credentials?

tmbo:
I mean there is already a mount for `endpoints` - they are in the same file. so we might consider renaming the `credentials.yml` to something more reasonable in the documentation, e.g. `endpoints.yml`

wochinge:
But `credentials.yml` somehow makes sense as it is specified by `--credentials`, doesn't it?

wochinge:
This is kind of escalating quickly 😄 but how about adding another volume so that users can add their project directory (training data, domain.yml, ...)? 
We could also make a volume `config` which can both contain `endpoints.yml` and `credentials.yml`. I think there is no real need to keep this in separate directories.

tmbo:
yes I agree, let's assume that these files are in the same directory

tmbo:
yes I agree, let's assume that these files are in the same directory

wochinge:
@tmbo ready for another review (commit might not yet be visible in Github due to Github's restore operations)

docker/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
+version: '3.0'
+
+services:
+  rasacore:
+      image: rasa/rasa_core:latest
+      networks: ['rasa-network']
+      ports:
+      - "5005:5005"
+      volumes:
+      - "./rasa-app-data/core-model:/app/model"
```
tmbo: we do have a starter pack to get run from when creating a new bot, can you please make sure the docker compose and that repository play well with one another https://github.com/RasaHQ/starter-pack-rasa-stack

docker/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
+version: '3.0'
+
+services:
+  rasacore:
```
tmbo: i think we should stick to one naming convention (`action_serve` vs `rasacore`), i guess this should be `rasa_core`

docker/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+        - --endpoints
+        - config/endpoints.yml
+  action_server:
+      image: rasa/rasa_core_sdk:latest
+      networks: ['rasa-network']
+      ports:
+      - "5055:5055"
+      volumes:
+      - "./rasa-app-data/actions:/app/actions"
+  rasanlu:
```
tmbo: `rasa_nlu`

docs/installation.rst
```diff
@@ -87,14 +90,82 @@ Rasa Stack starter-pack which has all the files you need to build your first cus
...
+
+.. note::
+
+    If you do not require components like `nlu <https://rasa.com/docs/nlu/>`_ or duckling,
+    you can simply remove them from your docker compose file.
+
+Running it
+~~~~~~~~~~
+To run all components locally, execute :code:`docker-compose up`.
+You can then interact with your chat bot using the `Rasa Core HTTP API <https://rasa.com/docs/core/server/>`_.
```
tmbo: since this is a link within the core docs, let's use 
```
:ref:`section_http` 
```

docs/installation.rst
```diff
@@ -87,14 +90,82 @@ Rasa Stack starter-pack which has all the files you need to build your first cus
...
+    .. code-block:: bash
+
+         docker-compose run rasacore train
+
+Volume Explanation
+~~~~~~~~~~~~~~~~~~
+- **./rasa-app-data/core-model**: This directory contains the trained Rasa Core models.
+  You can also move previously trained models to this directory to load them within the Docker container.
+- **./rasa-app-data/config**: This directory is for the configuration of the endpoints and of the
+  different `chat & voice platforms <https://rasa.com/docs/core/connectors/>`_ you can use Rasa Core with.
```
tmbo: same here, docs internal links should use `:ref:`

docs/installation.rst
```diff
@@ -87,14 +90,82 @@ Rasa Stack starter-pack which has all the files you need to build your first cus
...
+      For the shown docker-compose example the file should look like this:
+
+        .. code-block:: yaml
+
+            action_endpoint:
+                url: 'http://action_server:5055/webhook'
+            nlu:
+                url: 'http://rasanlu:5000'
+
+    - If you use a connector to a `chat & voice platforms <https://rasa.com/docs/core/connectors/>`_
```
tmbo: `:ref:`

PR #1202, opened 2018-10-21
http evaluation endpoint

### LINKED ISSUE #1201: http story evaluation endpoint

ricwo:
**Proposed changes**:
- endpoint for story evaluation at `/evaluate`
- option for e2e evaluation with `?e2e=true`
- evaluation returns new field `in_training_data_fraction`, which is the fraction of the evaluation's stories that were found in training data
- fixes #1201

- [x] updated the changelog


rasa_core/evaluate.py
```diff
@@ -324,13 +324,21 @@ def _predict_tracker_actions(tracker, agent, fail_on_prediction_errors=False,
...
         if isinstance(event, ActionExecuted):
-            action_executed_result = \
+            action_executed_result, policy = \
                 _collect_action_executed_predictions(
                         processor, partial_tracker, event,
                         fail_on_prediction_errors
                 )
+            if (test_in_training_data and
+                    policy is not None and
+                    "Memoization" not in policy):
```
tmbo: can we use `SimplePolicyEnsemble.is_not_memo_policy` instead? feel free to move the function somewhere else

rasa_core/evaluate.py
```diff
@@ -324,13 +324,21 @@ def _predict_tracker_actions(tracker, agent, fail_on_prediction_errors=False,
...
+
     for event in events[1:]:
         if isinstance(event, ActionExecuted):
-            action_executed_result = \
+            action_executed_result, policy = \
                 _collect_action_executed_predictions(
                         processor, partial_tracker, event,
                         fail_on_prediction_errors
                 )
+            if (test_in_training_data and
```
tmbo: I think we should put this whole logic into a separate function and not into `_predict_tracker_actions `. I think it is fine to iterate twice over the events

rasa_core/evaluate.py
```diff
@@ -378,12 +389,19 @@ def collect_story_predictions(
     report, precision, f1, accuracy = get_evaluation_metrics(
             [1] * len(completed_trackers), correct_dialogues)
 
+    try:
```
tmbo: same here, I think this whole logic should be separate

rasa_core/server.py
```diff
@@ -448,6 +453,27 @@ def load_model():
...
         return '', 204
 
+    @app.route("/evaluate",
+               methods=['POST', 'OPTIONS'])
+    @requires_auth(app, auth_token)
+    @cross_origin(origins=cors_origins)
+    def evaluate_stories():
+        """Evaluate stories against the currently loaded model."""
+
+        data = convert_bytes_to_string(request.get_data())
```
tmbo: `utils.convert_bytes_to_string` (we shouldn't import functions from somewhere, but always call them on the module they are in)

rasa_core/utils.py
```diff
@@ -588,6 +588,13 @@ def pad_list_to_size(_list, size, padding_value=None):
     return _list + [padding_value] * (size - len(_list))
 
 
+def convert_bytes_to_string(data):
```
tmbo: type annotation please

rasa_core/server.py
```diff
@@ -448,6 +453,27 @@ def load_model():
...
+        tmp = tempfile.NamedTemporaryFile(delete=False)
```
tmbo: I think we should use 
```
import rasa_nlu

rasa_nlu.utils.create_temporary_file(data)
```
in this case we also won't need the `convert_bytes_to_string`

tests/test_evaluation.py
```diff
@@ -17,7 +17,7 @@ def test_evaluation_image_creation(tmpdir, default_agent):
     stories_path = tmpdir.join("failed_stories.md").strpath
     img_path = tmpdir.join("evaluation.png").strpath
 
-    run_story_evaluation(
+    _ = run_story_evaluation(
```
tmbo: I don't think we should add this

tests/test_server.py
```diff
@@ -210,6 +212,32 @@ def test_predict(http_app, app):
...
+                                               "in_training_data_fraction"}
+
+
+def test_end_to_end_evaluation(app):
+    with io.open(END_TO_END_STORY_FILE, 'r') as f:
+        stories = f.read()
+    response = app.post('/evaluate?e2e=true',
+                        data=stories)
+    assert response.status_code == 200
+    assert set(response.get_json().keys()) == {"report",
```
tmbo: this isn't really testing if e2e is working - the response test is the same as the above

ricwo: Only way to know really is to add field to the return dict of `rasa_core.evaluate`

PR #1225, opened 2018-10-24
Evaluation improvements

ricwo:
**Proposed changes**:
- partial tracker uses correct user utterance and correct action, regardless of prediction
- entities are excluded from evaluation metrics
- the failed stories output prints user messages in end-to-end form, regardless of whether prediction was correct or not

- [ ] updated the changelog

### DIFF
--- a/rasa_core/evaluate.py
+++ b/rasa_core/evaluate.py
@@ -175,7 +175,36 @@ class WronglyPredictedAction(ActionExecuted):
                                                     self.predicted_action)
 
 
-class WronglyClassifiedUserUtterance(UserUttered):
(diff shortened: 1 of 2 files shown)


rasa_core/evaluate.py
```diff
@@ -164,17 +178,43 @@ class WronglyPredictedAction(ActionExecuted):
 
     type_name = "wrong_action"
 
-    def __init__(self, correct_action, predicted_action, timestamp=None):
```
tmbo: we should keep the timestamp, as it is an argument of the superclass

PR #1229, opened 2018-10-25
Add action_back to default actions

RolandJAAI:
**Proposed changes**:
- allow the user to go one step back in the conversation like "I would like to correct the last number"
- ability to offer a button "go one step back"
- the bot will utter the second last question again and wipe everything that happened afterwards from the tracker, the user can give a different answer
- very handy if the user gave a wrong answer before and wants to correct it
- also good if typos lead to wrong intent recognition without triggering fallback actions
- to reliably trigger the action_back it needs to be build-in like ActionRestart

This feature was highly demanded by multiple clients and our tests have proven it to be a great addition. However, the only way to trigger the action reliably in complex stories is by hard-wiring it like the restart action.

Behaviour of the action could be overwritten like the other default actions if needed. 

- [x] updated the changelog


RolandJAAI:
Closing and opening again to trigger new build as travis timed out

RolandJAAI:
@amn41 @akelad @tmbo Would be awesome if you could check this PR - being able to go one step back in a conversation from wherever you are has proven to be a very valuable feature for our clients. It makes the bot a lot more robust. Cheers, Roland

akelad:
Hi @RolandJAAI thanks for submitting this PR! We agree this is something that makes sense to add, but we're not entirely sure this is the best way to implement it yet. We'll discuss this and get back to you soon

RolandJAAI:
Thanks @akelad - glad you like the feature, I am open to do this any other way, for now I just tried to do it like the restart action is already implemented.

RolandJAAI:
Hi @akelad have you been able to discuss this? Like stated above, enabling the user to go one step back in a conversation from wherever he is whenever he wants to has proven to be a very valuable feature for our clients. It makes the bot a lot more robust. The current implementation is like for restart. 

In a next step a more general approach for reliably triggering any action by an intent could be implemented (maybe with an ActionPolicy which has a dict in the config, mapping intents to actions?)
Thanks, Roland

akelad:
@RolandJAAI Hi, sorry for getting back to you so late. Yes we discussed it and while we do like the functionality, we don't want to implement it like this. 
You're right with your suggestion for a more general implementation though, we'll implement a policy that handles this for any intents you define. I can't tell you exactly when this feature will come, but it's definitely something we're looking into

akelad:
This is the issue for that policy btw https://github.com/RasaHQ/rasa_core/issues/1596

PR #1242, opened 2018-10-29
Graphviz removal

wochinge:
**Proposed changes**:
- remove the (py)graphviz dependencies
- display the story graph within a html page instead of generating an image
- adapting the visualization feature of the interactive learning for that
- reuse the html template used for the visualization of the interactive learning
- adapt documentation etc to the new html approach

Command to test it out:
`python -m rasa_core.visualize -s examples/concertbot/data/stories.md -d examples/concertbot/domain.yml -o graph.html`

- [x] updated the changelog


amn41:
I think we need to add pydot to the setup.py as well.

amn41:
can confirm ff, safari still errors

wochinge:
I fixed the issue with safari. It was actually some problems with the replacement which made the js processing fail for interactive learning.
I also moved and adapted the test case.

tmbo:
works for me in safari, must be your old version @amn41 :D

Issue #1239, opened 2018-10-29
Docs on how to build a bot with rasa & docker


tmbo:
We do have some docks about the docker installation process, but there isn't a lot of information around the whole development process.

Idea: Have a walk through of the development with docker and the different steps involved in the creation of the chatbot:
- setup
- adding training data (nlu and core, creating a domain, mounting this)
- adding code (custom python code for actions using the sdk, maybe even including custom dependencies, so other python libraries that need to be installed )
- running it
- connecting it to a platform (facebook, ...)

tmbo:
@JustinaPetr do the above steps make sense, should we change the order?

tmbo:
I mean, I like to start from scratch, but I am not the model user for this - so better if @JustinaPetr takes a guess.

jonDel:
How about an example using a docker compose file, in order to run all the containers integrated as one service (rasa core, rasa nlu, sdk - actions, the tracker store, etc)?

JustinaPetr:
@tmbo @wochinge I think the steps above make sense. It's good to show the step-by-step process from scratch because it's important to enable anyone - from complete beginner to an experience Docker user to be able to follow.
Also, I think that starting from scratch is an easier way to explain the entire process step-by-step. So I guess reusing the data of the starterpack and the idea of the bot would still be ok, but I would go through the development process step-by-step from the very beginning.

PR #1254, opened 2018-10-31
add docker tutorial for core & nlu #1239

wochinge:
**Proposed changes**:
- see https://github.com/RasaHQ/rasa_core/issues/1239

- [x] add steps to set up core
- [x] add steps to set up core & nlu
- [x] instructions for custom pipeline
- [x] add steps for a custom action server
- [ ] add steps for duckling
- [x] add steps for custom tracker stores
- [x] updated the changelog

### DIFF
--- a/docs/docker_walkthrough.rst
+++ b/docs/docker_walkthrough.rst
@@ -4,23 +4,21 @@ Docker Walkthrough
 ==================
 
 This walkthrough provides a tutorial on how to set up Rasa Core, Rasa NLU,
-an Action Server, and Duckling with Docker containers. The tutorial starts from
-scratch. Hence, all you need is a bit of familiarity with your command line.
-
-Let's get started!
+and an Action Server with Docker containers.
+If you have not used Rasa before it is recommended to read the
+:ref:`quickstart`.
 
 .. contents::
 
 1. Setup
 --------
 
-In order to get going we need only two things installed.
+Requirements for the tutorial:
 
     - A text editor of your choice
     - Docker
 
-If you do not have installed `Docker <https://www.docker.com/>`_ so far, you
-have to install it, too. If you are not sure whether it is installed execute the
+If you are not sure whether Docker is installed on your machine execute the
 following command:
 
   .. code-block:: bash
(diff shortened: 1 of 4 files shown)


docs/docker_walkthrough.rst
```diff
@@ -0,0 +1,539 @@
...
+    docker -v && docker-compose -v
+    # Docker version 18.06.1-ce, build e68fc7a
+    # docker-compose version 1.22.0, build f46880f
+
+If your output is not similar to the one above, please install Docker.
+See `this instruction page <https://docs.docker.com/install/>`_
+of the Docker documentation which describes how to install Docker on a variety
+of operating systems.
+
+Congratulation! 🚀 Now we have got everything set up to get going!
```
JustinaPetr: The same with using 'we'. I would use 'you' everywhere because it refers to the user directly.  One more thing to this is that I would probably leave the 'Congratulations' part here because it makes it look like the most important part of the tutorial is already finished  :)

docs/docker_walkthrough.rst
```diff
@@ -0,0 +1,539 @@
...
+    - create some story lines for our chatbot
+    - train our chatbot
+    - run our chatbot within a Docker container
+
+2.1 Create Stories
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Rasa Core works by learning from example conversations. Therefore, we now need
+to write a few examples to get started.
+Stories are written in `Markdown <https://en.wikipedia.org/wiki/Markdown>`_
```
JustinaPetr: I would suggest having this part after you explain which directories a user will have to create because there is too much going back and forth. Maybe it would even make sense to create a separate step for preparing the directory and then diving into stories

docs/docker_walkthrough.rst
```diff
@@ -0,0 +1,539 @@
...
+  docker run \
+    -v $(pwd):/app/project \
+    -v $(pwd)/models/rasa_core:/app/models \
+    rasa/rasa_core:latest \
+    train \
+      --domain project/domain.yml \
+      --stories project/data/stories.md \
+      --out models
+
+Yey! You just trained your chatbot for the first time!
```
JustinaPetr: I think there are too many excitement words included :) For documentation, I would suggest keeping documentation content a bit more toned down because developers should use the docs as a concise and guide of how to use the software.

JustinaPetr: Also, I think it's better to leave out 'for the first time' part, because the quickstart is the place where we expect people to build their first Rasa assistant.

docs/docker_walkthrough.rst
```diff
@@ -0,0 +1,539 @@
...
+  ]
+
+If the bot cannot understand you (which is quite like since we provided only
+very little NLU training data), the answer is something along the lines:
+
+.. code-block:: bash
+
+  []
+
+Hooray! 🎉 🎊 We can now train Rasa Core, Rasa NLU and speak to our bot in
```
JustinaPetr: Should this really be about training? I think the training steps are already included above.

wochinge:
@JustinaPetr I implemented your comments and added the next steps.
It is still quite long, but I don't know where to shorten it without losing information.
I am also not sure about Duckling since I don't know how common it is to use Duckling as part of the NLU pipeline. Generally I tend to include in the tutorial as then everything is covered.

docs/docker_walkthrough.rst
```diff
@@ -0,0 +1,684 @@
...
+      command:
+        - start
+        - --path
+        - models
+        - -c
+        - config/nlu_config.yml
+
+.. note::
+
+  Depending on the selected pipeline you might have to use a different
```
JustinaPetr: Where a user can find more information about that? Would be good to add a link or something, because people do you their own custom configs very often.

docs/docker_walkthrough.rst
```diff
@@ -0,0 +1,684 @@
...
+    def name(self):
+      return "action_joke"
+
+    def run(self, dispatcher, tracker, domain):
+      request = requests.get('http://api.icndb.com/jokes/random').json() #make an api call
+      joke = request['value']['joke'] #extract a joke from returned json response
+      dispatcher.utter_message(joke) #send the message back to the user
+      return []
+
+Then add the custom action in your stories and your domain file.
```
JustinaPetr: 'Then' repeats two times in a row. What about alternating it with 'Next'? :)

docs/docker_walkthrough.rst
```diff
@@ -0,0 +1,684 @@
...
+Continuing the example from above replace replace ``utter_cheer_up`` in
```
JustinaPetr: Typo: 'replace' is written two times :)

docs/docker_walkthrough.rst
```diff
@@ -0,0 +1,684 @@
...
+
+As for Rasa NLU, it is necessary to tell Rasa Core the location of the action
+server. Add this to your ``config/endpoints.yml``:
+
+.. code-block:: yaml
+
+  action_endpoint:
+    url: http://action_server:5055/webhook
+
+If you then run ``docker-compose up`` the action server starts together
```
JustinaPetr: I would rephrase this as a mandatory step: 'Run 'docker-compose up' to start the action server together with...'.

JustinaPetr:
Awesome! I would be keen on leaving the duckling out just because it's very optional unless there are some special configurations which a developer would have to implement for the duckling to work. Is it very different from adding additional components and adding it to the config files?

wochinge:
@JustinaPetr Adding Duckling is just as the other components, except that the url of has to be added within the NLU pipeline

wochinge:
@JustinaPetr Updated it according to your notes

JustinaPetr:
@wochinge Looks good! :)

PR #1256, opened 2018-11-01
Custom store endpoints

wochinge:
**Goal**:
- Building on https://github.com/RasaHQ/rasa_core/pull/1048 the `endpoints.yml` should now also support the loading of custom written `TrackerStore`s. This is similar to what was done here: https://github.com/RasaHQ/rasa_core/pull/1169.  
- Issue: https://github.com/RasaHQ/rasa_core/issues/1199

**Proposed changes**:
- add possibility to load custom `TrackerStore`s using the module path
- non functional input leads to a warning and the use of the `InMemoryTrackerStore`
- assimilate `host` parameter of `RedisTrackerStore` and `MongoTrackerStore` to correspond to the
  field name in `endpoints.yml`
- add the required steps in the documentation

- [x] updated the changelog


wochinge:
@MetcalfeTom I am not sure how to deal with the `url` part.
The renaming / mapping we had so far (`url` to `host` for Redis and Mongo) adds no value and it also makes it more complicated to test the feature. 
So any suggestions on that (and all the rest) are appreciated 🙂

MetcalfeTom:
I agree with what you've implemented in that they should all be the same.  My suggestion is to rename the `endpoints.yml` field to `host` instead since it is more descriptive than `url`

wochinge:
Mhm, but I think a host is more likely a url, than url a host. Maybe I should remove the renaming part and simply make a local class inherit from the RedisTrackerStore for test purposes. How about that?

wochinge:
@MetcalfeTom I now created a custom tracker store for the tests and left the `url` `host` thing aside

PR #1271, opened 2018-11-03
Wrong HTTP response code when updating model from server

znat:
**Proposed changes**:
- small doc thing

- [ ] updated the changelog

### DIFF
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -88,7 +88,7 @@ Rasa Core sends requests to your model server with an ``If-None-Match``
 header that contains the current model hash. If your model server can
 provide a model with a different hash from the one you sent, it should send it
 in as a zip file with an ``ETag`` header containing the new hash. If not, Rasa
-Core expects an empty response with a ``204`` status code.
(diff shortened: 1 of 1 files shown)


akelad:
Thanks for submitting the PR! But I think the 204 is correct? https://github.com/RasaHQ/rasa_core/blob/master/rasa_core/agent.py#L139

tmbo:
@znat is right though, it should probably rather be 304 instead of 204 (I think we should change the code to accept both and treat them the same way)

znat:
204 is no content and 304 is not modified. I guess it depends on implementations so @tmbo's idea to support both is probably the best.

akelad:
@znat is that a modification you'd be willing to add to the code?

znat:
If you agree on the solution I can certainly submit a PR. I would also update the doc to mention both 204 and 304

akelad:
yes, sounds good!

znat:
Closing. See https://github.com/RasaHQ/rasa_core/pull/1307 instead

PR #1272, opened 2018-11-03
Fix update model from server

znat:
**Proposed changes**:
- Unzipping a model usually put the files in a default directory named after the archive name. If that's the case, domain and models should be taken from this directory (otherwise there is a FileNotFound error)

- [X ] made PR ready for code review
- [ ] updated the changelog

### DIFF
--- a/rasa_core/agent.py
+++ b/rasa_core/agent.py
@@ -66,6 +66,7 @@ def load_from_server(interpreter=None,  # type: NaturalLanguageInterpreter
                   action_endpoint=action_endpoint)
 
     if wait_time_between_pulls:
+        print ('start worker')
         # continuously pull the model every `wait_time_between_pulls` seconds
         start_model_pulling_in_worker(model_server,
                                       wait_time_between_pulls,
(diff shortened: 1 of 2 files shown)


akelad:

ricwo:
@znat thanks for submitting this PR! When you say _usually put the files in a default directory_, do you mean this also happens when the model is zipped from **within** the model directory? Because that's the intended zipping procedure before upload (`cd models/dialogue && zip -r core_model.zip *`)

ricwo:
@znat I see. But even if you zip the files in OS X and provide this zipped model at a model endpoint, the `zip_ref.extractall()` function handles it normally. The `pull_model_and_fingerprint` method isn't intended to look for local OSX-unzipped model files

znat:
>  But even if you zip the files in OS X and provide this zipped model at a model endpoint, the zip_ref.extractall() function handles it normally. 

The bug happens with a model zipped locally with the mac os utility, then uploaded to an endpoint. After fetching, core unzips in the tmp folder and looks for model in the same folder as the archive. But it's one level below in the folder created by the macos utility. So unless you don't expect users to zip model files manually and to only use a particular utility to build archive it is still an issue I think

ricwo:
@znat I cannot reproduce this bug - would you mind posting exact steps to reproduce?

When I use the osx-native tool (*right-click*->*Compress*) and then in a python shell run `zipfile.ZipFile('Archive.zip').extract_all('.')` on that archive, I get the zipped files themselves, _not_ contained in a default directory. See the screenshot attached - I presume you mean the same utility?

image

tmbo:
I'll close this, please let us know if there is still something we need to fix.

PR #1278, opened 2018-11-05
pass correct tracker store object #1277

wochinge:
**Proposed changes**:
- https://github.com/RasaHQ/rasa_core/issues/1277
- If no queue is present for the broker configuration, it overwrites the default queue with `None`. This should not be the case, the default value should be preserverd

- [x] updated the changelog

### DIFF
--- a/rasa_core/server.py
+++ b/rasa_core/server.py
@@ -19,10 +19,8 @@ from rasa_core import utils, constants
 from rasa_core.channels import CollectingOutputChannel, UserMessage
 from rasa_core.evaluate import run_story_evaluation
 from rasa_core.events import Event
-from rasa_core.interpreter import NaturalLanguageInterpreter
 from rasa_core.policies import PolicyEnsemble
 from rasa_core.trackers import DialogueStateTracker, EventVerbosity
-from rasa_core.utils import AvailableEndpoints
 from rasa_core.version import __version__
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


wochinge:
@tmbo 
- do we maintain`rasa_core/server.py`? Since at the moment `server.py` does not implement the new tracker store + broker feature
- do we need an update in the changelog

wochinge:
@tmbo updated

rasa_core/server.py
```diff
@@ -600,38 +598,5 @@ def tracker_predict():
...
-    run.serve_application(_agent,
-                          cmdline_args.connector,
-                          cmdline_args.port,
-                          cmdline_args.credentials,
-                          cmdline_args.cors,
-                          cmdline_args.auth_token,
-                          cmdline_args.enable_api,
-                          cmdline_args.jwt_secret,
-                          cmdline_args.jwt_method)
+    raise RuntimeError("Using `rasa_core.server` is not longer supported. "
```
tmbo:
```suggestion
    raise RuntimeError("Calling `rasa_core.server` directly is no longer supported. "
```

tmbo:
:D the build failed because of pep8 & the line I added being to long - very unfortunate that githubs code suggestion feature doesn't show line length :wink:

wochinge:
Yeah, I had this one before 😄 Fix is pushed.

PR #1281, opened 2018-11-06
Add mongo auth_source parameter

### LINKED ISSUE #1248: Add authSource argument to MongoClient

ricwo:
**Proposed changes**:
- fixes #1248 

- [x] updated the changelog


ricwo:
yes, good catch. `MongoClient` kwarg should be `authSource`: http://api.mongodb.com/python/current/api/pymongo/mongo_client.html

I don't think we can test this without running a mongo server during tests, which seems like overkill for this one parameter

MetcalfeTom:
You're right. I was going to suggest constructing a client locally and then using its class functions to test the `authSource` parameter but I can't find anything that uses it in the code.

PR #1283, opened 2018-11-07
Allow users to add environment variables in the configuration file. #1282

leon0707:
**Proposed changes**:
I created the issue #1282. This pr is suppose to fix it.

- allow users to add environment variables in the configuration file. 

- [x] updated the changelog


akelad:
Thanks for submitting this PR, I think this could be really useful! We'll give it a review as soon as possible

wochinge:
@leon0707 This is an awesome idea! 

I try it out later today, just a few things
- if we can do something like `${VARIABLE}/remaining` -> `valueOfVariable/remaining` it might be nice to also support `prefix/${VARIABLE}` -> `prefix/valueOfVariable`
- some tests would be great
  - environment variable present (no prefix, prefix, no remaining path, remaining path)
  - environment variable not present

rasa_core/utils.py
```diff
@@ -328,12 +328,39 @@ def construct_yaml_str(self, node):
...
+    def env_var_constructor(loader, node):
+        """Process environment variables found in the YAML."""
+        value = loader.construct_scalar(node)
+        envVar, remainingPath = env_var_pattern.match(value).groups()
+        return os.environ[envVar] + remainingPath
+
+    if six.PY2:
+        yaml.add_constructor(u'!env_var', env_var_constructor)
+    else:
+        yaml.constructor.SafeConstructor.add_constructor(
```
wochinge:
```suggestion
        yaml.SafeConstructor.add_constructor(
```

rasa_core/utils.py
```diff
@@ -328,12 +328,39 @@ def construct_yaml_str(self, node):
...
+    import os
+
+    # eg. ${USER_NAME}, ${PASSWORD}
+    env_var_pattern = re.compile(r"^\$\{(.*)\}(.*)$")
+    yaml.add_implicit_resolver('!env_var', env_var_pattern)
+
+    def env_var_constructor(loader, node):
+        """Process environment variables found in the YAML."""
+        value = loader.construct_scalar(node)
+        envVar, remainingPath = env_var_pattern.match(value).groups()
```
wochinge:
```suggestion
        env_var, remaining_path = env_var_pattern.match(value).groups()
```

rasa_core/utils.py
```diff
@@ -328,12 +328,39 @@ def construct_yaml_str(self, node):
...
+        return os.environ[envVar] + remainingPath
```
wochinge:
```suggestion
        return os.environ[env_var] + remaining_path
```

rasa_core/utils.py
```diff
@@ -328,12 +328,39 @@ def construct_yaml_str(self, node):
                                     construct_yaml_str)
 
 
+def process_env_var():
```
wochinge:
```suggestion
def replace_environment_variables():
```

rasa_core/utils.py
```diff
@@ -328,12 +328,39 @@ def construct_yaml_str(self, node):
...
+            u'!env_var', env_var_constructor)
+
+
 def read_yaml_file(filename):
     """Read contents of `filename` interpreting them as yaml."""
     return read_yaml_string(read_file(filename))
 
 
 def read_yaml_string(string):
+    process_env_var()
```
wochinge:
```suggestion
    replace_environment_variables()
```

wochinge:
So checked it out and played around a bit: Works like a charm 🚀 
I have a few suggestions on parameter naming, and the ones from the comment, but otherwise it's fine! Thanks for your contribution!

leon0707:
Thank you @CLAassistant @wochinge I will update the names and add test cases.

wochinge:
@leon0707 If you want so inspiration for the tests: Have a look at my pull request for NLU: https://github.com/RasaHQ/rasa_nlu/pull/1518

leon0707:
@wochinge I added a couple test cases. `prefix_${XXX}`, `prefix_${XXX}_postfix` and `${XXX}_postfix` are all supported.

tests/test_utils.py
```diff
@@ -118,3 +119,112 @@ def test_endpoint_config():
...
+         host: http://localhost:27017
+         db: rasa
+         user: ${USER_NAME}
+         password: ${PASSWORD}
+         collection: conversations
+    """
+    try:
+        r = utils.read_yaml_string(config_with_env_var)
+    except Exception as e:
+        assert isinstance(e, KeyError) and e.args[0] == 'PASSWORD'
```
wochinge: please use the pytest method to check for exceptions: https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

wochinge:
@leon0707 
Sorry, I think communicated the location of the environment variables  not clearly. What I meant

```python
os.environ['USER_NAME'] = 'user'
os.environ['PASS'] = 'pass'

def test_read_yaml_string()
   [...]

# other tests [...]
```
So put it **above** `test_read_yaml_string` but not in the methods.

One more thing, then we are good to go: Can you maybe trim the example configuration within your test cases, so that it is not that much duplicate code.  I mean the actual values are not important, we just want to test the correct replacement of the environment variables. Just have the minimal required config for your test case. For example instead of:

```
    config_with_env_var = """
    tracker_store:
      mongo:
         host: http://localhost:27017
         db: rasa
         user: ${USER_NAME}
         password: ${PASS}
         collection: conversations
    """
```

Just:

```python
    config_with_env_var = """
    mongo:
         user: ${USER_NAME}
         password: ${PASS}
    """
```

tests/test_utils.py
wochinge: @leon0707

CHANGELOG.rst
```diff
@@ -37,6 +37,9 @@ Added
 - add documentation on how to configure endpoints within a configuration file
 - ``auth_source`` parameter in ``MongoTrackerStore`` defining the database to
   authenticate against
+- environment variables specified with ``${env_variable}`` in a yaml
+    configuration file are now replaced with the value of the environment variable
+
```
```suggestion
```

PR #1289, opened 2018-11-07
Update channel.py

smn-snkl:
The button_to_string and send_custom_message function should not assume a specific structure of buttons.

E.g. I'm using some custom Facebook buttons (like the Login button) which don't have a payload or a subtitle. These two functions then throw an error when testing stories with custom buttons in the cmdline.

**Proposed changes**:
- Instead of directly accessing the assumed button properties, I propose using .get(attr, '') to fill in missing attributes with empty strings

Another option could be to just dump the element as a JSON string (`json.dumps(button, indent=2)`)

- [ ] updated the changelog


akelad:

PR #1298, opened 2018-11-08
add possibility to configure wait_time_between_pulls through the endp…

wochinge:
**Configure `wait_time_between_pulls` through the endpoint configuration file**

Issue: #1276

**Proposed changes**:
- remove parameter `wait_time_between_pulls` from `rasa_core.run` functions and `rasa_core.agent.load_from_server`
- use value from model endpoint configuration (use 100 as default value as before)

- [x] updated the changelog


wochinge:
My test cases are a bit weird with `monkeypatch`, but I don't want actually execute a request.

PR #1299, opened 2018-11-09
fix bug when logging failed e2e stories

yobekiko:
**Proposed changes**:
- Fix error when logging failed end-2-end stories during e2e evaluation

- [ ] updated the changelog

### DIFF
--- a/rasa_core/evaluate.py
+++ b/rasa_core/evaluate.py
@@ -204,7 +204,7 @@ class EndToEndUserUtterance(UserUttered):
     Mostly used to print the full end-to-end user message in the
     `failed_stories.md` output file."""
 
-    def as_story_string(self):
+    def as_story_string(self, e2e=True):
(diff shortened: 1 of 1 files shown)


akelad:
Thanks for raising this issue, we'll give it a review as soon as possible

rasa_core/evaluate.py
```diff
@@ -204,7 +204,7 @@ class EndToEndUserUtterance(UserUttered):
     Mostly used to print the full end-to-end user message in the
     `failed_stories.md` output file."""
 
-    def as_story_string(self):
+    def as_story_string(self, e2e=True):
```
ricwo: Is this kwarg necessary?

ricwo: I believe functionally it isn't but it preserves the signature of the parent method

yobekiko: Yes, exactly.

PR #1330, opened 2018-11-13
lots of small fixes for cli interaction

### LINKED ISSUE #1322: config is not passed to visualize, visualize report no FormPolicy for version 0.12.0
**Rasa Core version**: 0.12.0

**Python version**: 3.6

**Operating system** (windows, osx, ...):osx

**Issue**:
now policy is stored in config file while visualize.py doesn't use the -c argument. And default there is no form policy in visualize.py. When run visualize using domain with forms section. Error as below
```
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.6/site-packages/rasa_core/visualize.py", line 60, in <module>
    agent = Agent(args.domain, policies=[MemoizationPolicy(), KerasPolicy()])
  File "/usr/local/lib/python3.6/site-packages/rasa_core/agent.py", line 208, in __init__
    "You have defined a form action, but haven't added the "
rasa_core.domain.InvalidDomain: You have defined a form action, but haven't added the FormPolicy to your policy ensemble.
```

**Content of domain file** (if used & relevant):
```yaml

```

tmbo:
**Proposed changes**:
- fixes #1322
- allow to properly pass stories in the same way to all scripts
- use the same parsing behaviour of the policy config file for all scripts
- moved the default policy to the module, so that we can use it when rasa core gets packaged

- [x] updated the changelog


tmbo:
This is an important PR that we should merge for a 0.12.1 release to fix broken cmdline interaction (e.g. visualisation script)

rasa_core/train.py
```diff
@@ -212,47 +180,50 @@ def train_comparison_models(stories,
...
 def do_default_training(cmdline_args, stories, additional_arguments):
+    """Train a model."""
+
     if not cmdline_args.out:
         raise ValueError("you must provide a path where the model "
                          "will be saved using -o / --out")
+
     if (isinstance(cmdline_args.config, list) and
             len(cmdline_args.config) > 1):
         raise ValueError("You can only pass one config file at a time")
```
ricwo: We could do away with these checks by leaving `nargs` for `config` unspecified, in which case it is passed as a single string, not a list. The script won't even run if you'd specify more

rasa_core/cli/arguments.py
```diff
@@ -0,0 +1,55 @@
...
+import pkg_resources
+
+
+def add_config_arg(parser):
+    """Add an argument to the parser to request a policy configuration."""
+
+    parser.add_argument(
+            '-c', '--config',
+            type=str,
+            nargs="*",
```
ricwo: Can `nargs` be anything other than one (or the default)?

rasa_core/cli/arguments.py
```diff
@@ -0,0 +1,55 @@
...
+
+def add_domain_arg(parser):
+    """Add an argument to the parser to request a the domain file."""
+
+    parser.add_argument(
+            '-d', '--domain',
+            type=str,
+            required=True,
+            help="domain specification yaml file")
+    return parser
```
ricwo: Can we make these functions return `None` to indicate that `parser` is mutable? We don't use the return value in any of these functions' calls

rasa_core/config.py
```diff
@@ -11,10 +11,8 @@
 
 def load(config_file):
     # type: (Optional[Text], Dict[Text, Any], int) -> List[Policy]
```
ricwo: the `type` annotations don't seem right

rasa_core/evaluate.py
```diff
@@ -605,7 +599,7 @@ def run_comparison_evaluation(models, stories, output):
                                    num_correct)
 
 
-def plot_curve(output, no_stories, ax=None, **kwargs):
+def plot_curve(output, no_stories, ax=None):
     """Plot the results from run_comparison_evaluation."""
```
ricwo: Can we add type annotations and include the arguments in the description?

rasa_core/agent.py
```diff
@@ -611,11 +621,15 @@ def visualize(self,
         from rasa_core.training.dsl import StoryFileReader
         """Visualize the loaded training data from the resource."""
 
+        # if the user doesn't provide a max history, we will use the
+        # largest value from any policy
+        max_history = max_history or self._max_history()
```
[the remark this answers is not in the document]
tmbo: Yes, because no policy will have an exact recollection of what happened before max max steps. If we use the min, there would be policies with a longer recollection (e.g. memoization policy) that could differentiate between two conversations, so we should differentiate between them in the visualization as well.

rasa_core/train.py
tmbo: yes, that is a good point, at the moment we really do not need multiple configs for this script

tmbo:
Alright, done - ready for another review. I took your comment (adding description to parameters) as a reason to bring all the comments in the same google style docstr format.

rasa_core/channels/botframework.py
```diff
@@ -149,8 +149,9 @@ def __init__(self, app_id, app_password):
         # type: (Text, Text) -> None
         """Create a Bot Framework input channel.
 
-        :param app_id: Bot Framework's API id
-        :param app_password: Bot Framework application secret
+        Args:
+            app_id: Bot Framework's API id
+            app_password: Bot Framework application secret
```
ricwo: 🎢

PR #1344, opened 2018-11-14
Update button / element output in interactive training

smn-snkl:
Currently there's a few issues with outputting buttons and elements in interactive training causing issues like ([#1310](https://github.com/RasaHQ/rasa_core/issues/1310)):

- button_to_string: does not print anything except title and payload, so custom buttons won't have any output
- format_bot_output: does not add elements to output string

**Proposed changes**:
- Instead of only returning string with title and value of a button, I appended a json.dumps(button) for custom buttons
- For now I just added a json.dumps string of elements to the output string

The output after changes is not necessarily pretty, but more functional:
```
Hello!
Image: http://example.com/myimage.png
Attachment: My Attachment
Buttons:
1: yes (/yes) {"payload": "/yes", "title": "yes"}
2: no (/no) {"payload": "/no", "title": "no"}
Elements:
1: element1 {"buttons": [{"payload": "/button1", "title": "button1"}], "title": "element1"}
2: element2 {"buttons": [{"payload": "/button2", "title": "button2"}], "title": "element2"}
```
Open for prettier output options though :)

- [ ] updated the changelog

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -18,7 +18,7 @@ from rasa_core.featurizers import BinarySingleStateFeaturizer, \
 
 
 @pytest.mark.parametrize("filename", glob.glob(
-            "data/test_config/example_config.yaml"))
+    "data/test_config/example_config.yaml"))
 def test_load_config(filename):
     loaded = load(filename)
     assert len(loaded) == 2
@@ -61,7 +61,7 @@ def test_ensemble_from_dict():
     # Check if all policies are present
     assert len(ensemble) == 4
     # MemoizationPolicy is parent of FormPolicy
-    assert any([isinstance(p, MemoizationPolicy) and \
+    assert any([isinstance(p, MemoizationPolicy) and
                 not isinstance(p, FormPolicy) for p in ensemble])
     assert any([isinstance(p, KerasPolicy) for p in ensemble])
     assert any([isinstance(p, FallbackPolicy) for p in ensemble])


smn-snkl:
Instead of adding the entire json.dumps string, we could also pop title, payload, etc. first and then only add a json.dumps of the rest. So it would look more like this:
```
Buttons:
1: yes (/yes)
2: no (/no)
Elements:
1: element1 {"buttons": [{"payload": "/button1", "title": "button1"}]}
2: element2 {"buttons": [{"payload": "/button2", "title": "button2"}]}
```

akelad:
Hmm, that makes things a lot harder to read. I'm more for keeping the easy to read version

smn-snkl:
@akelad : The thing is that having for example a button like this (Messenger's Login Button):
```
{
    "type": "account_link",
    "url": FB_LOGIN_URL,
}
```
Doesn't output anything but the idx. So you don't know what the button is or does.

tmbo:
I actually quite like the second option - it is a little more concise but contains all the necessary details :+1: - do you mind adapting your PR to use your proposed format?

smn-snkl:
@akelad @tmbo: Accidentally force-pushed to this patch. Fixed everything now and updated the `patch-3` PR branch to the proposed format, but the PR is closed now. Should I open a new one?

Also added the new button + element representations to the console channel in console.py

akelad:
Cool, thanks! We'll give it a review

PR #1347, opened 2018-11-14
Ensemble.from_dict not working

smn-snkl:
- get_featurizer_from_dict and get_state_featurizer_from_dict are called like classmethods in from_dict, but classmethod decorator is missing.
- Instead of get_state_featurizer_from_dict, get_featurizer_from_dict is called twice

**Proposed changes**:
- Added `@classmethod` to both methods
- Changed get_featurizer_from_dict to get_state_featurizer_from_dict when state featurizer is created

- [ ] updated the changelog

### DIFF
--- a/rasa_core/policies/ensemble.py
+++ b/rasa_core/policies/ensemble.py
@@ -208,7 +208,8 @@ class PolicyEnsemble(object):
 
                 if featurizer_config.get('state_featurizer'):
                     state_featurizer_func, state_featurizer_config = \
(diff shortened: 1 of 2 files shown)


smn-snkl:
This should also fix this: [#1341](https://github.com/RasaHQ/rasa_core/issues/1341)

akelad:
Thanks for this quick fix! Do you mind adding to your test the check for the presence of the featurizers and state featurizers as well?

smn-snkl:
Of course :)

PR #1369, opened 2018-11-19
fix an error during interactive training with actions using utter_cus…

wochinge:
**Issue**:
- https://github.com/RasaHQ/rasa_core/issues/1080

**Proposed changes**:
- always have empty output string even if there is no text in bots message
- safe access of buttons key in channels 

- [x] updated the changelog

### DIFF
--- a/rasa_core/training/interactive.py
+++ b/rasa_core/training/interactive.py
@@ -239,10 +239,8 @@ def format_bot_output(
 ) -> Text:
     """Format a bot response to be displayed in the history table."""
 
-    if "text" in message:
-        output = message.get("text")
-    else:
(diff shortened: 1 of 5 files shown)


rasa_core/training/interactive.py
```diff
@@ -239,10 +239,8 @@ def format_bot_output(
 ) -> Text:
     """Format a bot response to be displayed in the history table."""
 
-    if "text" in message:
-        output = message.get("text")
-    else:
-        output = ""
+    output = message.get("text") or ""
+    output = "" if output is None else output
```
wochinge:
```suggestion
```

PR #1353, opened 2018-11-15
test domain save from interactive training

tzolkincz:
**Proposed changes**:
- There is a bug with domain persistence from interactive learning. When domain is saved some default actions (which are not overridden) are added to the domain config file.
Scope of this PR is only to add test, to show buggy behavior and start discussion how this issue can be resolved.

**Status (please check what you already did)**:
- [x] made PR ready for code review
- [x] added some tests for the functionality
- [ ] updated the documentation
- [ ] updated the changelog

PR #1377, opened 2018-11-21
test domain save from interactive training - merge

tzolkincz:
**Proposed changes**:
- PR for merge as discussed in #1353 (I've deleted source branch and now cannot reopen the original PR)

- [ ] updated the changelog

### DIFF
--- a/tests/test_interactive.py
+++ b/tests/test_interactive.py
@@ -226,4 +226,3 @@ def test_interactive_domain_persistance(mock_endpoint, tmpdir):
 
     httpretty.disable()
     assert "action_listen" not in saved_domain["actions"]
-
--- a/tests/test_interactive.py
+++ b/tests/test_interactive.py
@@ -227,3 +227,4 @@ def test_interactive_domain_persistance(mock_endpoint, tmpdir):
 
     httpretty.disable()
     assert "action_listen" not in saved_domain["actions"]
+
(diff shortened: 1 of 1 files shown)


Ghostvv:
Great! thanks, now let's see if test is passing

Ghostvv:
For some reason other than py 3.6 tests do not pass (we're looking into it), but your test seems to be passing now. Could you please add assert that other default actions are not present as well

tzolkincz:
Sure, I'll add other asserts as well.

tests/test_interactive.py
```diff
@@ -6,6 +6,7 @@
 from rasa_core import utils
 from rasa_core.training import interactive
 from rasa_core.utils import EndpointConfig
+from rasa_core.actions import action
```
Ghostvv: let's import the function
```suggestion
from rasa_core.actions.action import default_actions
```

tests/test_interactive.py
```diff
@@ -205,3 +206,25 @@ def test_undo_latest_msg(mock_endpoint):
...
+
+    events = tracker_json.get("events", [])
+
+    domain_path = tmpdir.join("interactive_domain_save.yml")
+    interactive._write_domain_to_file(domain_path, events, mock_endpoint)
+
+    saved_domain = utils.read_yaml_file(domain_path)
+
+    httpretty.disable()
+    for default_action in action.default_actions():
```
Ghostvv:
```suggestion
    for default_action in default_actions():
```

PR #1378, opened 2018-11-21
remove keras requirment since we moved to tf.keras

Ghostvv:
**Proposed changes**:
- remove keras requirment since we moved to tf.keras

- [x] updated the changelog

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,6 @@ numpy==1.14.5
 typing==3.6.4
 ruamel.yaml==0.15.66
 requests==2.20.0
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -8,7 +8,6 @@ numpy==1.14.5
 typing==3.6.4
 ruamel.yaml==0.15.66
 requests==2.20.0
-keras==2.1.6
```
tmbo: also needs to be removed from `setup.py`

Ghostvv: done

PR #1379, opened 2018-11-21
beginning

znat:
**Proposed changes**:
- Work in progress (not ready to merge)
Those are just minimal changes to make it work with the Rasa Webchat with session persistence (https://github.com/mrbot-ai/rasa-webchat/tree/feature/session-persistence)
To import the bundle directly in a `<script>` tag: https://storage.googleapis.com/mrbot-cdn/webchat-feature/session-persistence.js

The problem with this implementation is that sending a message to a `recipient_id` outside of the context of an incoming message (e.g. broadcasting) won't work because the `sid` is stored in the output channel when the message is received and used when the output channel replies.

I think the only way to overcome that limitation is to keep a map of session ids and socket ids, which then implies the same persistence problems as the tracker store (in case of a multi nodes setup or simply if core restarts)

- [ ] updated the changelog

### DIFF
--- a/rasa_core/channels/socketio.py
+++ b/rasa_core/channels/socketio.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import logging
+import uuid
 from typing import Optional, Text
 
 import socketio
@@ -24,8 +25,7 @@ class SocketBlueprint(Blueprint):
         super(SocketBlueprint, self).__init__(*args, **kwargs)
 
     def register(self, app, options, first_registration=False):
-        app.wsgi_app = socketio.Middleware(self.sio, app.wsgi_app,
-                                           self.socketio_path)
+        app.wsgi_app = socketio.Middleware(self.sio, app.wsgi_app, self.socketio_path)
         super(SocketBlueprint, self).register(app, options, first_registration)
 
 
@@ -35,25 +35,22 @@ class SocketIOOutput(OutputChannel):
     def name(cls):
         return "socketio"
 
-    def __init__(self, sio, bot_message_evt):
+    def __init__(self, sio, sid, bot_message_evt):
         self.sio = sio
+        self.sid = sid
         self.bot_message_evt = bot_message_evt
 
-    def send(self, recipient_id, message):
-        # type: (Text, Any) -> None
-        """Sends a message to the recipient."""
-        self.sio.emit(message, room=recipient_id)
-
-    def _send_message(self, recipient_id, response):
+    def _send_message(self, socket_id, response):
         # type: (Text, Any) -> None
         """Sends a message to the recipient using the bot event."""
-        self.sio.emit(self.bot_message_evt, response, room=recipient_id)
+        print (response)
+        self.sio.emit(self.bot_message_evt, response, room=socket_id)
 
     def send_text_message(self, recipient_id, message):
         # type: (Text, Text) -> None
         """Send a message through this channel."""
 
-        self._send_message(recipient_id, {"text": message})
+        self._send_message(self.sid, {"text": message})
 
     def send_image_url(self, recipient_id, image_url):
         # type: (Text, Text) -> None
@@ -64,7 +61,7 @@ class SocketIOOutput(OutputChannel):
                 "payload": {"src": image_url}
             }
         }
-        self._send_message(recipient_id, message)
+        self._send_message(self.sid, message)
 
     def send_text_with_buttons(self, recipient_id, text, buttons, **kwargs):
         # type: (Text, Text, List[Dict[Text, Any]], **Any) -> None
@@ -82,7 +79,7 @@ class SocketIOOutput(OutputChannel):
                 "payload": button['payload']
             })
 
-        self._send_message(recipient_id, message)
+        self._send_message(self.sid, message)
 
     def send_custom_message(self, recipient_id, elements):
         # type: (Text, List[Dict[Text, Any]]) -> None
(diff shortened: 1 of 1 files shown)


tmbo:
@znat not sure yet I completely understand what you are trying to achieve. What do you mean by broadcasting? can you describe an example use case where you would use that?

tmbo:
The explicit session generation that is part of this PR looks like a really good idea - just want to understand your question better.

znat:
Hi @tmbo, have you discussed with @akela? The idea is to enable session persistence on the rasa-webchat. 
I'm not trying to broadcast or anything, I am just pointing out that this implementation is incomplete because it ignores the `recipient_id` param on the `utter_...` functions. And that the only way I see to avoid this is a SessionStore mapping socket ids and core sessions ids. But then, it becomes a bigger thing as we have to worry about the persistence of this SessionStore. Does that make more sense?

znat:
So my question is more: how do you want to handle it? Leave it as is it, go for the Session Store idea (with the risk of having to support many implementations like the tracker stores), or any better idea?

tmbo:
I think you made the right choice: The sender_id is more to help the channel create a connection back to the user, if it is not necessary because the connection is based on some other information from the input channel that is fine.

So the idea of the implementation is good, and I'd keep it that way avoiding an id store for the socket IO connections.

znat:
> if it is not necessary because the connection is based on some other information from the input channel that is fine.
If we accept the idea that it is not possible to recreate a connection back to the user with this channel for now (which would prevent handoff I guess), there is still the problem of a misleading API. Because `utter_...` functions still have this param exposed.

rasa_core/channels/socketio.py
```diff
@@ -35,25 +35,22 @@ class SocketIOOutput(OutputChannel):
...
-        # type: (Text, Any) -> None
-        """Sends a message to the recipient."""
-        self.sio.emit(message, room=recipient_id)
-
-    def _send_message(self, recipient_id, response):
+    def _send_message(self, socket_id, response):
         # type: (Text, Any) -> None
         """Sends a message to the recipient using the bot event."""
-        self.sio.emit(self.bot_message_evt, response, room=recipient_id)
+        print (response)
```
tmbo:
```suggestion

```

rasa_core/channels/socketio.py
```diff
@@ -125,10 +122,9 @@ def __init__(self,
 
     def blueprint(self, on_new_message):
         sio = socketio.Server()
-        socketio_webhook = SocketBlueprint(sio, self.socketio_path,
-                                           'socketio_webhook', __name__)
+        socketio_webhook = SocketBlueprint(sio, self.socketio_path, 'socketio_webhook', __name__)
 
-        @socketio_webhook.route("/", methods=['GET'])
+        @socketio_webhook.route("/health", methods=['GET'])
```
tmbo: the `health` endpoints for all the other channels is at `/` as well, do we need to change this here?

rasa_core/channels/socketio.py
```diff
@@ -141,11 +137,24 @@ def disconnect(sid):
             logger.debug("User {} disconnected from socketio endpoint."
                          "".format(sid))
 
+        @sio.on('session_request', namespace=self.namespace)
+        def session_request(sid, data):
+            if data is None:
+                data = {}
+            if 'session_id' not in data or data['session_id'] is None:
+                data['session_id'] = str(uuid.uuid4())
```
tmbo:
```suggestion
                data['session_id'] = uuid.uuid4().hex
```

rasa_core/channels/socketio.py
```diff
@@ -141,11 +137,24 @@ def disconnect(sid):
...
+            sio.emit("session_confirm", data['session_id'])
+            logger.debug("User {} disconnected from socketio endpoint."
```
tmbo: I guess we mean "connected"?
```suggestion
            logger.debug("User {} connected to socketio endpoint."
```

rasa_core/channels/socketio.py
```diff
@@ -141,11 +137,24 @@ def disconnect(sid):
...
         def handle_message(sid, data):
-            output_channel = SocketIOOutput(sio, self.bot_message_evt)
-            message = UserMessage(data['message'], output_channel, sid,
-                                  input_channel=self.name())
-            on_new_message(message)
-
-        return socketio_webhook
+            output_channel = SocketIOOutput(sio, sid, self.bot_message_evt)
+            if "session_id" not in data or data["session_id"] is None:
+                logger.warning("A message without a valid sender_id was received")
```
tmbo: For this case, can we not use the sid as we did before?

rasa_core/channels/socketio.py
```diff
@@ -141,11 +137,24 @@ def disconnect(sid):
...
+            else:
+                message = UserMessage(data['message'], output_channel, data['session_id'],
+                                      input_channel=self.name())
+                on_new_message(message)
+
+        return socketio_webhook
```
tmbo: please add a newline

rasa_core/channels/socketio.py
znat: To avoid breaking changes I added a flag in the credentials and put a condition here.

akelad:
@znat thanks for getting this done so quick! There's a few pep8 checks that failed, could you take a look at them?

akelad:
@znat is this good to merge apart from the pep8 fixes?

znat:
@akelad Yes, should be. I don't really understand the issue raised by code climate though

akelad:
it's not codeclimate, but the travis build throwing pep8 errors: https://travis-ci.com/RasaHQ/rasa_core/jobs/162269985#L1180
Sorry if that wasn't clear

znat:
Right!
I am really trying but pep8 is never happy... And it's really hard to make sense of it without a VGA monitor...

PR #1386, opened 2018-11-22
In `format_bot_output`, consider case where data key is present on dict, but None

m90:
When [passing an event](https://github.com/RasaHQ/rasa_core/blob/0.12.2/rasa_core/training/interactive.py#L522) like BotUttered("only a text message") to format_bot_output this will throw as [the dict being created](https://github.com/RasaHQ/rasa_core_sdk/blob/47106d959767148d15b328a3a38a95cd697e0616/rasa_core_sdk/events.py#L26-L33) will contain a data key, but its value is None. This requires an additional check for this case before proceeding as `get` might as well return `None`.

- [ ] updated the changelog

### DIFF
--- a/rasa_core/training/interactive.py
+++ b/rasa_core/training/interactive.py
@@ -241,8 +241,11 @@ def format_bot_output(
 
     output = message.get("text") or ""
 
-    # Append all additional items
+    # Append all additional items if data is not None
     data = message.get("data", {})
+    if not data:
+        return output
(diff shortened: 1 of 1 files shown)


m90:
Travis fails because of trailing whitespace, but for reasons unknown to me I currently cannot clone from GitHub and had to use the online editor. I will fix that once things are back to normal.

Issue #1390, opened 2018-11-23
TrainingDataGenerator#generate hangs infinitely


m90:
**Rasa Core version**: 0.12.2 (also can repro against master)

**Python version**: 3.6

**Operating system** (windows, osx, ...): Ubuntu 18.04

**Issue**:

I am migrating an existing Rasa bot from 0.11 to 0.12. The bot itself works just as before, yet I am running into one issue when trying to do interactive training: The `TrainingDataGenerator` created by `training.load_data` before running the interactive learning I/0 will hang infinitely on `generate` being called here:  https://github.com/RasaHQ/rasa_core/blob/c82d9107021d7d681f4beb60b7dbdb7ff60a7031/rasa_core/training/__init__.py#L57

When I look at the logging output I can see that the analysis assumes it has found unused checkpoints and does another iteration. Yet, this iteration will still not finish and at the 3rd or 4th loop my computer will usually freeze:

```
2018-11-23 20:20:08 DEBUG    rasa_core.training.generator  - Starting data generation round 0 ... (with 1 trackers)
Processed Story Blocks: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 198/198 [00:37<00:00,  5.25it/s, # trackers=2]
2018-11-23 20:20:46 DEBUG    rasa_core.training.generator  - Finished phase (66670 training samples found).
2018-11-23 20:20:46 DEBUG    rasa_core.training.generator  - Found 3 unused checkpoints in current phase.
2018-11-23 20:20:46 DEBUG    rasa_core.training.generator  - Found 1280 active trackers for these checkpoints.
2018-11-23 20:20:46 DEBUG    rasa_core.training.generator  - Starting data generation round 1 ... (with 1280 trackers)
Processed Story Blocks: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 198/198 [01:21<00:00,  2.43it/s, # trackers=8192]
2018-11-23 20:22:08 DEBUG    rasa_core.training.generator  - Finished phase (199790 training samples found).
2018-11-23 20:22:08 DEBUG    rasa_core.training.generator  - Found 29 unused checkpoints in current phase.
2018-11-23 20:22:08 DEBUG    rasa_core.training.generator  - Found 121344 active trackers for these checkpoints.
2018-11-23 20:22:08 DEBUG    rasa_core.training.generator  - Starting data generation round 2 ... (with 121344 trackers)
```

... and so on

[...]

PR #1410, opened 2018-11-28
Fix output of training with --debug_plots

m90:
This picks up #1390. I tried using `--debug_plots` but ran into two issues:

- a pdf file is exported, but it seems to be a HTML document
- when opening this document in a browser it will not display anything, throwing:

```
dagre-d3.min.js:26 Uncaught TypeError: shape is not a function
    at SVGGElement.<anonymous> (dagre-d3.min.js:26)
    at Selection.selection_each [as each] (d3.v4.js:1185)
    at createNodes (dagre-d3.min.js:26)
    at fn (dagre-d3.min.js:90)
    at Selection.selection_call [as call] (d3.v4.js:1149)
    at drawGraph (story_blocks_connections.html:114)
    at story_blocks_connections.html:916
```

I managed to fix both issues and can use the patched version as expected.

**Proposed changes**:
- Change the file extension of `debug_plots` so that it matches the output (it's a HTML file, not a PDF)
- Use a shape type supported by `dagre-d3`, see https://github.com/dagrejs/dagre-d3/issues/131

- [ ] updated the changelog

### DIFF
--- a/rasa_core/train.py
+++ b/rasa_core/train.py
@@ -154,7 +154,7 @@ def add_general_args(parser):
         action='store_true',
         help="If enabled, will create plots showing checkpoints "
              "and their connections between story blocks in a  "
-             "file called `story_blocks_connections.pdf`.")
+             "file called `story_blocks_connections.html`.")
 
     utils.add_logging_option_arguments(parser)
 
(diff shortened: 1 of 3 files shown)


akelad:
thanks for submitting this PR, we'll give it a review as soon as possible

m90:
@Ghostvv Done so!

PR #1417, opened 2018-11-30
Do not override custom button type if provided (#1416)

shasha79:
**Proposed changes**:
- Currently, Rasa overrides custom buttons to be always of the type "postback" even if the type is different and being provided. Changed this behavior to allow other button types (web_url etc.)
- fixes #1417 

- [ ] updated the changelog


akelad:

PR #1425, opened 2018-12-02
Add documentation for Policy.load and Policy.persist

andrewluotechnologies:
For implementing custom Policy classes

It'd be useful to add these methods to the API documentation - to use Agent.load and Agent.persist with a custom policy, you have to implement these methods in your custom Policy, but this isn't documented anywhere yet


akelad:

PR #1446, opened 2018-12-05
Pycodestyle

wochinge:
**Proposed changes**:
- fixes https://github.com/RasaHQ/rasa_core/issues/1445

- ~[ ] added some tests for the functionality~
- [x] updated the changelog

### DIFF
--- a/rasa_core/policies/tf_utils.py
+++ b/rasa_core/policies/tf_utils.py
@@ -223,8 +223,8 @@ def _compute_time_attention(attention_mechanism, attn_inputs, attention_state,
 # noinspection PyProtectedMember
 class TimeAttentionWrapperState(
     namedtuple("TimeAttentionWrapperState",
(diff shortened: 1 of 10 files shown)


rasa_core/policies/tf_utils.py
```diff
@@ -223,8 +223,8 @@ def _compute_time_attention(attention_mechanism, attn_inputs, attention_state,
 # noinspection PyProtectedMember
 class TimeAttentionWrapperState(
     namedtuple("TimeAttentionWrapperState",
-               tf.contrib.seq2seq.AttentionWrapperState._fields +
```
tmbo: can you please revert this file and put it into the ignored list for the style check?

This is copied and slightly modified from tensorflow and to make it easier to upgrade this in the feature it is better to not modify anything we do not need to modify here.

wochinge: alright!

PR #1461, opened 2018-12-07
Update facebook channel: handle message with an attachment

leon0707:
**Proposed changes**:
fix: #1460
- facebook messenger channel can handle messages with attachments.

- [ ] updated the changelog


akelad:

rasa_core/channels/facebook.py
```diff
@@ -50,6 +57,9 @@ def message(self, message: Dict[Text, Any]) -> None:
         elif self._is_audio_message(message):
```
wochinge: Since your `is_attachment_message` covers the `is_audio_message`, we can remove `_is_audio_message` and only use your implementation.

rasa_core/channels/facebook.py
```diff
@@ -50,6 +57,9 @@ def message(self, message: Dict[Text, Any]) -> None:
...
             attachment = message['message']['attachments'][0]
             text = attachment['payload']['url']
+        elif self._is_attachment_message(message):
+            attachment = message['message']['attachments'][0]
```
wochinge: what do you think of the idea to support multiple attachments? We could split the different urls by a newline in the code. E.g.

```python
attachments = []
for attachment in message['message'].get('attachments', []):
    url = attachment.get('payload', {}).get('url')
    attachments.append(url)

text = '\n'.join(attachments)
```

tmbo:
Sorry, this has been closed by accident.

rasa_core/channels/facebook.py
leon0707: I'm not sure it's safe to do that. Maybe we should state that this function is deprecated.

rasa_core/channels/facebook.py
leon0707: As far as I remember, I have tested it on facebook messenger(weeks ago BTY). I typed some texts and added 2 attachments into one message, then sent it out.  The backend actually got 3 messages separately. Each message only has one type of content. One for text and other two for attachments.

rasa_core/channels/facebook.py
wochinge: I checked and it's the same behavior: it adds the url of the attachment as text. So you can safely drop the audio part.

rasa_core/channels/facebook.py
wochinge: 👍 Cool, that you tried it out. Is it the same if you send multiple images?

wochinge:
@leon0707 How is it going with the pr? Do you need help?

PR #1490, opened 2018-12-17
Adding capability to load multi domain yml files from a folder 

ggdupont:
Simply listing the yml files for the input folder, lading them one by one and merging the Domain objects into agent self.domain.

Tested locally with few domain.yml (few thousands lines each) with no apparent issue. That's far from "real tests" but I did not find testing guidelines in the contribution section and don't really know the python best practices... Feel free to suggest something so I can improve the PR.

**Proposed changes**:
- see agent.py starting line 190

- [ ] updated the changelog


akelad:

ggdupont:
> Thanks for raising this PR, we'll give it a review as soon as possible

On my end, I'm checking if I can sign the CLA on my name (not involving my employer).
On the CI part, not sure I understood the errors raised by Travis...

rgstephens:
Rather than providing a list of domain.yml files, I think it would be better for the `-d` option to accept the name of a directory which contains a collection of files. This is more consistent with the `-s` stories option which accepts a directory of files.

MetcalfeTom:
@ggdupont Nothing major - Travis does some pep8 checks which your code failed on.  Fixing simply involves [removing whitespaces from these lines](https://travis-ci.com/RasaHQ/rasa_core/jobs/165647479#L1167)

Please update the `CHANGELOG.rst` and write a test - we have a folder called `test_domains` under `/data` which should work fine for this.  I don't think a docs update is necessary

After that, good to go!

tmbo:
We'll need to close this for the time being.

We did quite some work on a new commandline interface to run the scripts which includes loading the files. Based on that, I don't think it makes sense to load domain files from the same folder. What we rather expect is to load multiple specified domain files from different folders. Since the merging functionality is already implemented, the call to merge needs to be somewhere else though.

Thanks a lot for the work you have put into the PR, but I don't think this change makes sense anymore.

ggdupont:
Arf... I'm too slow.

Just to clarify, you meant the work on the CLI naturally made available such feature? (ie I have a use-case where I dynamically load a dialog agent not from CLI)

tmbo:
@ggdupont not necessarily tied to CLI, I mean in the end the argparsers call python functions that you can use as well :wink:

The merging of domain files isn't implemented there yet either - so still a todo, but the code and structure moved which is why the PR doesn't make sense anymore.

PR #1422, opened 2018-11-30
Validate slot

ksurya:
**Proposed changes**:
- Related to https://github.com/RasaHQ/rasa_core_sdk/pull/28

**Status (please check what you already did)**:
- [x] made PR ready for code review
- [ ] added some tests for the functionality
- [x] updated the documentation
- [ ] updated the changelog

PR #1506, opened 2018-12-20
update formbot example

JustinaPetr:
**Proposed changes**:
Since this example will be referenced in the forms tutorial, I made some changes to improve the conversation part of this example.

- Added intent greet and added some training data for it
- Changed entity annotation for numbers (numbers in a word format are picked up by duckling only so having an annotation for num_people didn't make much sense)
- Slightly changed the utterance templates to make the conversation a bit smoother
- Updated the Makefile to use -c policy.yml

- [ ] updated the changelog

### DIFF
--- a/examples/formbot/domain.yml
+++ b/examples/formbot/domain.yml
@@ -8,10 +8,10 @@ intents:
   - deny
   - stop
   - thankyou
+  - greet
 
 entities:
   - cuisine
-  - num_people
   - number
   - feedback
   - seating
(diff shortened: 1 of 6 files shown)


examples/formbot/domain.yml
```diff
@@ -20,7 +20,7 @@ slots:
   cuisine:
     type: unfeaturized
     auto_fill: false
-  num_people:
+  number:
```
Ghostvv: I think we specifically created slot `num_people` to show how fill slot with different name from the entity

wochinge:
@Juste Can you please also update the missing Python3 type annotations in `actions.py`?

Ghostvv:
There is another PR to update formbot example: https://github.com/RasaHQ/rasa_core/pull/1422

PR #1521, opened 2018-12-27
fix error with e2e and unseen entities

EPedrotti:
**Proposed changes**:
- check if entity is correctly populated with the "entity" key, if not skips unrecognised entity

- [ ] updated the changelog


rasa_core/domain.py
```diff
@@ -409,8 +409,9 @@ def get_parsing_states(self,
             intent_config = self.intent_config(intent_name)
             should_use_entity = intent_config.get('use_entities', True)
             if should_use_entity:
-                key = "entity_{0}".format(entity["entity"])
-                state_dict[key] = 1.0
+                if "entity" in entity.keys():
```
ricwo:
```suggestion
                if "entity" in entity:
```

tests/test_evaluation.py
```diff
@@ -5,6 +5,7 @@
     run_story_evaluation,
     collect_story_predictions)
 from tests.conftest import DEFAULT_STORIES_FILE, END_TO_END_STORY_FILE
+from tests.conftest import E2E_STORY_FILE_UNKNOWN_ENTITY
```
ricwo: maybe import this on the same line as the two files above

tests/test_evaluation.py
```diff
@@ -50,3 +51,18 @@ def test_end_to_end_evaluation_script(tmpdir, default_agent):
         has_prediction_target_mismatch()
     assert len(story_evaluation.failed_stories) == 0
     assert num_stories == 2
+
+
+def test_end_to_end_evaluation_script_unknow_entity(tmpdir, default_agent):
```
ricwo:
```suggestion
def test_end_to_end_evaluation_script_unknown_entity(tmpdir, default_agent):
```

rasa_core/domain.py
ricwo: probably more readable if we put the two `if`s on the same line with an `and`

data/test_evaluations/story_unknown_entity.md
```diff
@@ -0,0 +1,6 @@
+
```
ricwo: do we need this blank line?

PR #1528, opened 2019-01-02
Add tf and sklearn version info to policy metadata

### LINKED ISSUE #1463: Extend model metadata to include tf and sklearn versions

ricwo:
**Proposed changes**:
- fixes #1463 

- [x] updated the changelog

### DIFF
--- a/rasa_core/policies/ensemble.py
+++ b/rasa_core/policies/ensemble.py
@@ -1,12 +1,12 @@
-import sys
-from collections import defaultdict
-
 import json
 import logging
-import numpy as np
 import os
+import sys
+from collections import defaultdict
 from datetime import datetime
 from typing import Text, Optional, Any, List, Dict, Tuple
+
+import numpy as np
 
 import rasa_core
 from rasa_core import utils, training, constants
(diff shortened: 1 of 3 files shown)


rasa_core/policies/ensemble.py
```diff
@@ -97,12 +99,22 @@ def _create_action_fingerprints(training_events):
             action_fingerprints[k] = {"slots": slots}
         return action_fingerprints
 
+    def _add_module_version_info(self, metadata: Dict[Text, Any]) -> None:
+        """Adds tensorflow and sklearn version info to metadata."""
+        if any(isinstance(p, KerasPolicy) for p in self.policies):
```
tmbo: I think instead of checking for the policy name, we could also just try to import, e.g.
```python
try:
    import tensorflow as tf
    metadata["tensorflow"] = tf.__version__
except ImportError:
    pass
```

We could even make this more general, create a class variable `versioned_packages = ["tensorflow", "sklearn", "rasa_core"]` and then iterate over that:
```python
import importlib
for package_name in self.versioned_packages:
    p = importlib.import_module(package_name)
    metadata[package_name] = p.__version__
```

ricwo: Wouldn't we only want the version info in the metadata if it was actually used during training?

tmbo: This is a good point, this depends where the check for compatibility will be implemented. But I guess that is going to be in the policy used, so there shouldn't be an issue with persisting the verison even if tf is not used. 

The reason why I don't want to use the KerasPolicy class check is, that e.g. the `EmbeddingPolicy` is using TF as well and is not inheriting from `KerasPolicy` so we would not persist the version.

ricwo: 👍 In that case it's ready for another review

rasa_core/policies/ensemble.py
```diff
@@ -97,12 +100,21 @@ def _create_action_fingerprints(training_events):
             action_fingerprints[k] = {"slots": slots}
         return action_fingerprints
 
+    def _add_package_version_info(self, metadata: Dict[Text, Any]) -> None:
+        """Adds version info for self.versioned_packages to metadata."""
```
```suggestion
        """Adds version info for self.versioned_packages to metadata."""

```

Issue #1561, opened 2019-01-11
[New feature] Try all the NLU predicted intents


oniz93:
I just developed a little change to processor.py
The change is the following:
- user input something
- if the predicted action is "action_default_fallback", try to use the next predicted intent until all the predicted intents are left.

I developed this change because my stories are like a tree so I need to understand which branch of the tree the user want to go even when the best predicted intent isn't in one of possible next stories.

I hope I'm explained well and I can give better informations or examples.

If it's an interesting feature for you, I'll make a pull request :)

akelad:
Hmm, I'm not sure I understand the need for this -- it will fall back on NLU confidence for all those intents since the first intent already trigerred it

oniz93:
I'll try to explain better.

I have a stories like this:

* greetings
- "choose the contest"
* contest_1 (or other blocks with contest_2, contest_3, etc.)
- "tell me you question"
* question_1_about_the_contest (or other blocks with other questions)

If a user, while choosing the contest, write something that is secognized as question, but with less confidence, can even be the sentence that triggers the contest selection, I'll iterate all the prediction until a valid story.
If I finish all the predictions "action_default_fallback" is triggered.

Let me know if I have explained well because english isn't my native language.

akelad:
So this is about the Core confidence, rather than NLU confidence right? I'll discuss this with the team, but i'm not sure whether it makes sense to implement this

wochinge:
So you basically try to force Core to follow a path of your training data, right?
I think there are currently better options to do that
* use the Keras Policy with a featurizer which considers the intent probabilities
* https://rasa.com/docs/core/master/policies/#two-stage-fallback-policy

oniz93:
@wochinge I tried to use Keras Policy but isn't the best option for my case.
Keras policy tries to figure which could be the next action based on the intent probabilites but can't follow a story graph structured like a tree.

Just to do an example, if you see my story above:
- user choose contest 1
(Rasa follow the path "contest 1")
- user say a question included in contest 2 (but with a lower probability can be a question for contest 1)
if I use Keras Policy: Rasa will recognize the intent as a question included in another story and will use that action (with a wrong answare so).
if I don't use Keras Policy: Rasa will throw "fallback" because the most probable intent isn't in the story

That's why I want to iterate the other probable intents.

Two Stage Fallback Policy isn't good for me because I have a really probable intent recognized by NLU (over 0.8 confidence).

wochinge:
[...]

PR #1562, opened 2019-01-11
Code for Issue #1561.

oniz93:
I bit more extended message is in the Issue #1561 

**Proposed changes**:
- New feature that test all the NLU's predicted intents instead of directly throwing "action_default_fallback"

- [ ] updated the changelog

### DIFF
--- a/rasa_core/processor.py
+++ b/rasa_core/processor.py
@@ -269,8 +269,9 @@ class MessageProcessor(object):
         intent_name = parse_data["intent"]
 
         # if isn't the first time that I enter in this function
-        # pop the intent_ranking of the predicted intents and use it for the next action prediction
-        if len(parse_data["intent_ranking"]) > 0 and firstTime == False:
+        # pop the intent_ranking of the predicted intents
+        # and use it for the next action prediction
+        if len(parse_data["intent_ranking"]) > 0 and firstTime is False:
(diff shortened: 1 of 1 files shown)


akelad:
Thanks for raising this PR, as mentioned we'll discuss this and get back to you

wochinge:
Hi @oniz93 ,
thanks for your effort and your time to this pull request 🚀 🙏 
However, after discussing your approach in our team, we decided to not integrate this into Rasa Core.
This was due to these reasons:
* If the NLU intent was not recognised correctly, this seems to be more a NLU problem and should therefore fixed by adding / changing the NLU training data
* If multiple intents might appear at a certain point in a story, you can use [OR statements](https://rasa.com/docs/core/stories/#or-statements)
* This feels like a too brute force way to solve other problems (see point 1)

Although we decided to close this pull request, I'd be very happy to see further contributions of yours to Rasa and maybe you can even publish this as a pypi package.

One more thing regarding the implementation:
I think you could also overwrite `action_default_fallback` and use that to revert the last user intent, and add a new UserMessageEvent with the 2nd entry in the intent ranking. In case the 2nd also leads to another `action_default_fallback` you could again revert the events, and try the 3rd ranked intent and so on.

Thanks for your support!

PR #1571, opened 2019-01-14
Random Seed for KerasPolicy and EmbeddingPolicy

RolandJAAI:
**Proposed changes**:
- As discussed in https://github.com/RasaHQ/rasa_nlu/pull/1509 this PR introduces a random seed for the KerasPolicy
- Default is None
- Setting the random_seed attribute to any integer will make the training results reproducible for the same inputs which can make a huge difference in production with regards to the fallback thresholds but also the action prediction itself can sometimes give surprising results without a fixed random seed.

In order for this to work we needed to make the sorting order of the story steps reproducible so that they are also fed into the NN in the same order in each run. Thanks to @guanjianyu for doing much of the hard work to figure out how it can be done. 

Once this is merged we can do the same for the embedding policy with just a few lines of code, too.

- [x] updated the changelog


akelad:

Ghostvv:
@RolandJAAI thank you for remembering to add seed to `core` as well. Do you mind adding it to `embedding_policy`?

Ghostvv:
Do you mind merging master?

docs/policies.rst
```diff
@@ -216,6 +216,10 @@ and the training is run here:
 You can implement the model of your choice by overriding these methods,
 or initialize ``KerasPolicy`` with pre-defined ``keras model``.
 
+In order to get reproducible training results for the same inputs you can
+set the ``random_seed`` attribute of the ``KerasPolicy`` to any integer. If
+the results are not good, try setting to another integer.
```
Ghostvv: Let's rephrase it. Because the main reason for bad results is probably not the random seed

docs/policies.rst
```diff
@@ -324,6 +328,9 @@ It is recommended to use
             - ``epochs`` sets the number of times the algorithm will see
               training data, where ``one epoch`` = one forward pass and
               one backward pass of all the training examples;
+            - ``random_seed`` if set to any int will get reproducible
+              training results for the same inputs, if the results are not
+              good, try setting to another integer;
```
Ghostvv: Let's rephrase it. Because the main reason for bad results is probably not the random seed

rasa_core/policies/embedding_policy.py
```diff
@@ -74,6 +74,9 @@ class EmbeddingPolicy(Policy):
         "batch_size": [8, 32],
         # number of epochs
         "epochs": 1,
+        # set random seed to any int to get reproducible results
+        # try to change to another int if you are not getting good results
```
Ghostvv: Let's rephrase it. Because the main reason for bad results is probably not the random seed

rasa_core/policies/embedding_policy.py
```diff
@@ -948,6 +953,10 @@ def train(self,
         self.graph = tf.Graph()
 
         with self.graph.as_default():
+            # set random seed
+            np.random.seed(self.random_seed)
+            tf.set_random_seed(self.random_seed)
```
Ghostvv: let's do them in the beginning of `train` in case we add some operations there

rasa_core/policies/keras_policy.py
```diff
@@ -138,12 +142,16 @@ def train(self,
         training_data = self.featurize_for_training(training_trackers,
                                                     domain,
                                                     **kwargs)
-
+        # set random seed for shuffle
+        import numpy as np
+        np.random.seed(self.random_seed)
```
Ghostvv: let's do them in the beginning of `train` in case we add some operations there

rasa_core/policies/keras_policy.py
```diff
@@ -138,12 +142,16 @@ def train(self,
...
         # noinspection PyPep8Naming
         shuffled_X, shuffled_y = training_data.shuffled_X_y()
 
         self.graph = tf.Graph()
         with self.graph.as_default():
+            # set random seed in tf
+            tf.set_random_seed(self.random_seed)
```
Ghostvv: let's do them in the beginning of `train` in case we add some operations there

rasa_core/policies/keras_policy.py
```diff
@@ -138,12 +142,16 @@ def train(self,
         training_data = self.featurize_for_training(training_trackers,
                                                     domain,
                                                     **kwargs)
-
+        # set random seed for shuffle
+        import numpy as np
```
Ghostvv: let's import `numpy` together with other imports

PR #1583, opened 2019-01-16
colored user input on cmdline chat

### LINKED ISSUE #1577: Set colors in console output
**Rasa Core version: 0.11.12**

**Python version: 3.6**

**Operating system (windows, osx, ...): Docker image python 3.6, with Linux Kernel.**

**Issue:**

When starting a conversation with the bot through the terminal, it is extremely relevant to have good visibility about what is asked to the bot, as well as what is returned to the user as a response of the interaction.

This issue is to develop a way to differentiate the messages that are sent to the bot from the messages that are issued by the user, applying the use of a different color from the existing one to the messages that are sent to the bot.

Is there a specific color or more suitable for the realization of it?

tmbo:
**Proposed changes**:
- fixes #1577

This is how it looks:
bildschirmfoto 2019-01-16 um 14 32 48

@JustinaPetr any thoughts on this?

- [ ] updated the changelog

### DIFF
--- a/rasa_core/channels/console.py
+++ b/rasa_core/channels/console.py
@@ -1,8 +1,9 @@
 # this builtin is needed so we can overwrite in test
-from builtins import input
+import questionary
 
 import json
 import requests
+from prompt_toolkit.styles import Style
 
 from rasa_core import utils
 from rasa_core.channels import UserMessage
@@ -34,7 +35,8 @@ def print_bot_output(message, color=utils.bcolors.OKBLUE):
 
 
 def get_cmd_input():
-    return input().strip()
+    return questionary.text("", style=Style([('qmark', '#b373d6'),
+                                             ('', '#b373d6')])).ask().strip()
 
 
 def send_message_receive_block(server_url, auth_token, sender_id, message):


JustinaPetr:
I think it looks good. Much easier to see what message was sent by the user.

The only thing, which is not exactly related to the user messages, is that warnings stand out so much imo, but I can't think of a better color as red or orange would only make things worse.

Ghostvv:
can we also get rid of this:
```
2019-01-18 12:15:55 WARNING  py.warnings  - /usr/local/lib/python3.6/site-packages/pykwalify/core.py:99: UnsafeLoaderWarning: 
The default 'Loader' for 'load(stream)' without further arguments can be unsafe.
Use 'load(stream, Loader=ruamel.yaml.Loader)' explicitly if that is OK.
Alternatively include the following in your code:

  import warnings
  warnings.simplefilter('ignore', ruamel.yaml.error.UnsafeLoaderWarning)

In most other cases you should consider using 'safe_load(stream)'
  data = yaml.load(stream)
```

Ghostvv:
amazing 😍 are you sure about question mark though? maybe `->`?

tmbo:
yes we can get rid of the warning, I think the most recent update of the library fixes this.

tmbo:
sure, we can also use an arrow, @JustinaPetr what do you think is the least confusing prefix for the input line?

rasa_core/channels/console.py
```diff
@@ -34,7 +35,8 @@ def print_bot_output(message, color=utils.bcolors.OKBLUE):
 
 
 def get_cmd_input():
-    return input().strip()
+    return questionary.text("", style=Style([('qmark', '#b373d6'),
```
Ghostvv: I would use an arrow here

JustinaPetr:
Regarding the prefix, the most natural would probably a colon, but not sure if that would be intuitive enough without any text like 'Your input:'.  I think the arrow or simply a dash would work pretty well.

tmbo:
I mean we can also put a short text there if you want. Let me combine that, and lets go for `Your input -> ` :D

PR #1591, opened 2019-01-17
fix fallback

### LINKED ISSUE #1533: FallbackPolicy shouldn't act on NLU confidence if the last user message was part of a form.
**Rasa Core version**: 0.12.3

**Rasa NLU version**: 0.13.8

**Rasa Core SDK version**: 0.12.1

**Python version**: 3.6.7

**Operating system** (windows, osx, ...): Linux Ubuntu 18.04

**Issue**: Hi, over the last few weeks I have been trying to find a way to find a way to handle generic user messages. In particular, in some specific cases, I would like the system to bypass intent recognition and entity extraction on what the user says, and just go on with the conversation according to the stories. Unfortunately I did not manage that, since the fallback action is always called if no intent is recognized with enough confidence, and the conversation is broken. So far I have temporarily solved the problem by adding some NLU examples of generic user sentences, but this can't really be a solution, since too many examples would be needed, and it would also create confusion when it comes to recognize different intents. Is there a way to just let the user say whatever he likes, possibly just save his sentence in a DB by means of an action and go on with the conversation regardless of the intent and entities?

Here follows, for clarity, an example of how I would like my stories to look like:

```markdown
## My story
> some_checkpoint
    - utter_question_1
* generic_user_input
    - action_save_user_response_in_db
    - utter_question_2
* generic_user_input
    - action_save_user_response_in_db
    ...
```

I have looked around and found something similar in other issues, but I didn't really understand how the solution would work. 

Thank you in advance for your help, let me know if more information is needed.

Ghostvv:
**Proposed changes**:
- change condition to check for latest action to be listen
- fixes #1533 

- [x] updated the changelog

### DIFF
--- a/rasa_core/policies/fallback.py
+++ b/rasa_core/policies/fallback.py
@@ -105,9 +105,7 @@ class FallbackPolicy(Policy):
                                       tracker.latest_action_name):
             logger.debug("NLU confidence {} is lower "
                          "than NLU threshold {}. "
-                         "Predicting fallback action: {}"
-                         "".format(nlu_confidence, self.nlu_threshold,
-                                   self.fallback_action_name))
+                         "".format(nlu_confidence, self.nlu_threshold))
             # we set this to 1.1 to make sure fallback overrides
             # the memoization policy
             result = self.fallback_scores(domain)
--- a/rasa_core/policies/fallback.py
+++ b/rasa_core/policies/fallback.py
@@ -85,7 +85,7 @@ class FallbackPolicy(Policy):
                                      domain: Domain) -> List[float]:
         """Predicts a fallback action.
 
-        Fallback predicted if NLU confidence is low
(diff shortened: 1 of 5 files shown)


rasa_core/policies/fallback.py
```diff
@@ -80,8 +83,11 @@ def fallback_scores(self, domain, fallback_score=FALLBACK_SCORE):
     def predict_action_probabilities(self,
                                      tracker: DialogueStateTracker,
                                      domain: Domain) -> List[float]:
-        """Predicts a fallback action if NLU confidence is low
-            or no other policy has a high-confidence prediction"""
+        """Predicts a fallback action.
+
+        Fallback predicted if NLU confidence is low
```
wochinge:
```suggestion
        The fallback action is predicted if the NLU confidence is low
```

PR #1604, opened 2019-01-21
Logging of low core confidence when FallbackAction is triggered

DavidSted:
**Proposed changes**:
- This PR adds some rasa_core logging information.
- The core model did already log the confidence for an intent prediction if the nlu_threshold wasn't reached. Now the core model (the SimplePolicyEnsemble) will also log its highest confidence for the next action to predict if a FallbackAction would be triggered because the core confidence threshold is not reached.

In a future PR we would like to add logging information in the core_evaluation on the core_confidence as it was already done for the nlu_evaluation.

- [ ] updated the changelog


akelad:
Thanks for submitting this PR, we'll give it a review as soon as possible. In the meantime, do you mind signing the CLA?

rasa_core/policies/ensemble.py
```diff
@@ -347,10 +351,14 @@ def probabilities_using_best_policy(self,
...
                 result = probabilities
+                best_policy = p
                 best_policy_name = 'policy_{}_{}'.format(i, type(p).__name__)
                 best_policy_priority = p.priority
+            if not isinstance(p, FallbackPolicy) and confidence > \
+                    max_confidence_no_fallback:
+                max_confidence_no_fallback = confidence
 
-        if (result.index(max_confidence) ==
+            if (result.index(max_confidence) ==
```
```suggestion
        if (result.index(max_confidence) ==
```

erohmensing:
Hi @DavidSted, I appreciate that you're improving our logging! Though, I'm not sure this is the simplest way to implement what you're trying to achieve. Is your main goal here basically to be notified when the FallbackPolicy is triggered because of low core confidence (instead of low NLU confidence)? Do we get that much out of knowing the highest confidence for another action that didn't reach the threshold, or are we just providing that information to parallel the current logging of low NLU confidence?

DavidSted:
> Hi @DavidSted, I appreciate that you're improving our logging! Though, I'm not sure this is the simplest way to implement what you're trying to achieve. Is your main goal here basically to be notified when the FallbackPolicy is triggered because of low core confidence (instead of low NLU confidence)? Do we get that much out of knowing the highest confidence for another action that didn't reach the threshold, or are we just providing that information to parallel the current logging of low NLU confidence?

Yes, the goal is to be notified when the FallbackPolicy is triggered because of low core confidence (instead of low NLU confidence). This could definitely be implemented more easier, but in addition the goal is to log the highest confidence for another action that didn't reach the threshold in that situation.

The reason or use case for that is that we want to know how insecure the core model was, when predicting a fallback policy. This information can give insight on how to change the training data to improve the whole model performance. 

So in the end the goal is to log some information like this: "Highest (non fallback) core confidence X is below threshold Y. Predicting fallback action: Z"

erohmensing:
Can you elaborate on this:

> The reason or use case for that is that we want to know how insecure the core model was, when predicting a fallback policy. This information can give insight on how to change the training data to improve the whole model performance.

Keeping in mind that the highest confidence that is below a core threshold will only come from a machine-learning policy (usually KerasPolicy), not the MemoizationPolicy or FormPolicy (as they both return confidence 0 or 1)?

DavidSted:
Well I think, a low core confidence of a machine-learning policy for a next action, even if the story is in the training data, might indicate that the model wasn't able to learn the desired behavior and the stories should eventually be changed.

Another utility of this logging information would be to see from a high confidence which is still below the core threshold, whether one just has to adjust the core confidence threshold so that more correct predictions of a machine-learning policy will be accepted.

erohmensing:
> Another utility of this logging information would be to see from a high confidence which is still below the core threshold, whether one just has to adjust the core confidence threshold so that more correct predictions of a machine-learning policy will be accepted.

I somewhat agree with your point here, although I think in order to know if you should lower your threshold, you have to also know the predicted next action, in order to know if it's correct or not. 


As for your PR, I really appreciate the effort, but I don't think how we want to go about adding this information. I do think some logging in the case of low core confidence is a worthwhile addition, though, and will make sure that some gets added.

PR #1613, opened 2019-01-21
Better creation and handling of facebook buttons

akelad:
So far quick reply buttons (e.g. for Facebook) can't really be used efficiently with rasa_core. Also Facebook buttons are forced to type postback, e.g. preventing Login buttons.

Proposed changes:

* Added respective string representation for quick replies to channel.py, interactive.py and console.py similar to the button_to_string function
* Changed the _add_postback_info function in facebook.py to only set button type to postback if button type is not specified by user
Changes made by @smn-snkl 

- [x] updated the changelog


akelad:
@smn-snkl did you test whether this works? Because I'm just getting error messages from the fbmessenger library. If yes, how did you define the utterances in your domain file?

smn-snkl:
@akelad : I think I did back then. Are you sure this PR is right? I see you're trying to merge patch-2 into master. But in the [original PR](https://github.com/RasaHQ/rasa_core/pull/1471/files) it's patch-1.

smn-snkl:
Oh, sorry. I mixed up these two PRs. Let me check what's going on here.

smn-snkl:
@akelad : Just verified it's working for me. I'm specifying the buttons according to the Messenger API ([this](https://developers.facebook.com/docs/messenger-platform/send-messages/quick-replies/))

For example, this is what I have in my domain file:

```
utter_help.quick_replies:
  - text: "How can I help?"
    quick_replies:
      - content_type: "text"
        title: "Search"
        payload: "/search"
```

smn-snkl:
@akelad : Two more things though. 

1. I think it would be useful to modify this section here in the dispatcher to include the quick_replies as well:
```
def utter_response(self, message: Dict[Text, Any]) -> None:
        """Send a message to the client."""

        bot_message = BotMessage(text=message.get("text"),
                                 data={"elements": message.get("elements"),
                                       "buttons": message.get("buttons"),
                                       "quick_replies": message.get("quick_replies"),  # does this work?
                                       "attachment": message.get("image")})

        self.latest_bot_messages.append(bot_message)
        self.output_channel.send_response(self.sender_id, message)
```

2. If you think it would be useful, we might want to rename `content_type` to just `type` to be more generic and avoid confusion with other platforms

What do you think?

akelad:
@tmbo what do you think about adding that stuff to the dispatcher? Not sure if it's not a bit too facebook specific

smn-snkl:
Maybe we can think of a more generic naming. Most platforms however, support some form of reply suggestions, e.g. Facebook's `quick replies`, Google's `suggestion chips`, Telegram's `Reply Keyboard Markup`

tmbo:
Thanks a lot for the added facebook feature :+1: - I'll merge it as it is now, we'll need to refactor the input/output channels a bit more in the coming weeks and we can come up with a better name then

PR #1619, opened 2019-01-22
Recover the slots values from database

Lucioric2000:
Nowadays, when a conversation is retrieved using MongoTrackerStore, the slots are returned empty. I propose this change to be able to recover the stored slot values.

**Proposed changes**:
- ...

- [ ** ] made PR ready for code review
- [ ] updated the changelog


akelad:
Thanks for submitting this PR, we'll give it a review as soon as possible. In the meantime, can you please rebase this off of master?

tmbo:
I don't think this change is necessary, the slot values should be recovered from the stored `SlotSet` events which are part of `stored.get("events")`. Did you have any issues with that?

Lucioric2000:
My slots are initialized as empty, although the values in the 'slots' field of the conversation document in MongoDB has fields with non-empty values.
using the versions of software:
rasa-core==0.12.1
rasa-core-sdk==0.12.1
rasa-nlu==0.13.7

tmbo:
This is not a proper solution to the issue.

All the events (e.g. `SlotSet`) are stored in mongo. When these events are replayed when the tracker is loaded, the slots will be set properly. If you modify the slots without going through an event, then it is intentional that the values will be lost.

PR #1621, opened 2019-01-23
handle passing in None for a channels route

tmbo:
**Proposed changes**:
- properly handle if `route` is set to `None` for a blueprint of an input channel when registering

- [ ] updated the changelog

### DIFF
--- a/tests/test_channels.py
+++ b/tests/test_channels.py
@@ -756,14 +756,12 @@ def test_newsline_strip():
 def test_register_channel_without_route():
     """Check we properly connect the input channel blueprint if route is None"""
     from rasa_core.channels import RestInput
+    from flask import Flask
+    import rasa_core
 
     # load your trained agent
     agent = Agent.load(MODEL_PATH, interpreter=RegexInterpreter())
-
     input_channel = RestInput()
-
-    from flask import Flask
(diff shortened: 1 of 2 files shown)


tests/test_channels.py
```diff
@@ -747,3 +751,26 @@ def test_newsline_strip():
...
+    """Check we properly connect the input channel blueprint if route is None"""
+    from rasa_core.channels import RestInput
+
+    # load your trained agent
+    agent = Agent.load(MODEL_PATH, interpreter=RegexInterpreter())
+
+    input_channel = RestInput()
+
+    from flask import Flask
+    import rasa_core
```
MetcalfeTom: Is this entire import necessary?

tmbo: yes, for `rasa_core.channels.channel.register`

MetcalfeTom: I was able to test this one locally by importing `channel` and then using `channel.register`

tmbo: yes sure, but that doesn't make any difference (ok it makes the line a little shorter where we call `register` but thats it :D - given that this is the way we call it in the other channels as well I don't see a reason to change this)

PR #1651, opened 2019-01-29
WIP: log duplicate stories

wochinge:
**Proposed changes**:
- log duplicate stories

Things to do:
- [ ] check with checkpointed stories
- [ ] check if check for phase is required and potentially remove it

- [ ] updated the changelog


wochinge:
@tmbo To be honest, I am not a 100 percent sure what happens in the code around it. Currently every duplicate story gets logged twice. Any ideas how to avoid that?

tmbo (approved):
I think the code looks good, not sure though where the double logging comes from - maybe @Ghostvv has some more insights on that

Ghostvv:
@wochinge didn't we try it together, and there were no double logging?

wochinge:
@Ghostvv Yes, there was no double logging with you. But still have to test this out with checkpointed stories and other stuff is currently more important. Should also be easy to be ported.

PR #1679, opened 2019-02-04
Fill form slots from form trigger message intent

smn-snkl:
@Ghostvv and myself discussed a way to allow form slots to be filled from the intent of the trigger message [here](https://forum.rasa.com/t/fill-form-slot-with-intent-of-initial-message/4744)

**Proposed changes**:
- Form event is now initialized with its trigger message
- Added a `trigger_message` attribute to `tracker.active_form`

- [x] updated the changelog

### DIFF
--- a/rasa_core/events/__init__.py
+++ b/rasa_core/events/__init__.py
@@ -835,8 +835,9 @@ class Form(Event):
     """
     type_name = "form"
 
-    def __init__(self, name, timestamp=None):
+    def __init__(self, name, trigger_message=None, timestamp=None):
         self.name = name
+        self.trigger_message = trigger_message if trigger_message else {}
         super(Form, self).__init__(timestamp)
 
     def __str__(self):
@@ -858,16 +859,18 @@ class Form(Event):
     @classmethod
     def _from_story_string(cls, parameters):
         """Called to convert a parsed story line into an event."""
-        return [Form(parameters.get("name"),
-                     parameters.get("timestamp"))]
+        return [Form(
+            name=parameters.get("name"),
+            trigger_message=parameters.get("trigger_message"),
+            timestamp=parameters.get("timestamp"))]
 
     def as_dict(self):
         d = super(Form, self).as_dict()
-        d.update({"name": self.name})
+        d.update({"name": self.name, "trigger_message": self.trigger_message})
         return d
 
     def apply_to(self, tracker: 'DialogueStateTracker') -> None:
-        tracker.change_form_to(self.name)
+        tracker.change_form_to(self.name, self.trigger_message)
 
 
 class FormValidation(Event):
--- a/rasa_core/trackers.py
+++ b/rasa_core/trackers.py
@@ -145,12 +145,13 @@ class DialogueStateTracker(object):
         generated_states = domain.states_for_tracker_history(self)
         return deque((frozenset(s.items()) for s in generated_states))
 
-    def change_form_to(self, form_name: Text) -> None:
+    def change_form_to(self, form_name: Text, trigger_message: Text) -> None:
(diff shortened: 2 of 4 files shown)


smn-snkl:
Respective changes in `rasa_core_sdk` can be found [here](https://github.com/RasaHQ/rasa_core_sdk/pull/44)

smn-snkl:
@Ghostvv : Any idea why the `trigger_message` is stored in the tracker when running core in regular mode, but not in `interactive` mode?

Ghostvv:
@smn-snkl thanks a lot for this work. I was on holiday, I'll review it this week

rasa_core/trackers.py
```diff
@@ -145,12 +145,13 @@ def past_states(self, domain: 'Domain') -> deque:
         generated_states = domain.states_for_tracker_history(self)
         return deque((frozenset(s.items()) for s in generated_states))
 
-    def change_form_to(self, form_name: Text) -> None:
+    def change_form_to(self, form_name: Text, trigger_message: Text) -> None:
```
Ghostvv: I don't think `trigger_message` is a `Text` here

Ghostvv: Actually I think adding trigger message to the `Form` event is unnecessary. At this point `'trigger_message': self.latest_message`, no?

rasa_core/events/__init__.py
```diff
@@ -835,8 +835,9 @@ class Form(Event):
     """
     type_name = "form"
 
-    def __init__(self, name, timestamp=None):
+    def __init__(self, name, trigger_message=None, timestamp=None):
```
Ghostvv: `trigger_message` is unnecessary here. See other comment

smn-snkl:
@Ghostvv : Yea, I think so. That's what I initially meant on the [forum](https://forum.rasa.com/t/fill-form-slot-with-intent-of-initial-message/4744/36?u=smn-snkl). 

There we had
```
def change_form_to(self, form_name: Text) -> None:
    """Activate or deactivate a form"""
    if form_name is not None:
        self.active_form = {'name': form_name,
                            'validate': True,
                            'rejected': False,
                            'trigger_message': self.latest_message}
    else:
        self.active_form = {}
```
I think `self.latest_message` is enough though, right? Then you have access to all the relevant data

Ghostvv:
sorry, I didn't get it from our discussion. Yes, but better put `self.latest_message.parse_data` directly so that is serializable. Like it is done here: https://github.com/RasaHQ/rasa_core/blob/7451990c9d3948709e4dc59f05f18d8be77e1a45/rasa_core/trackers.py#L132

smn-snkl:
Ah, yes you're right. Sorry, I got `rasa_core` mixed up with `rasa_core_sdk` where `tracker.latest_message` is already the parsed one.

```
# latest_message is `parse_data`,
# which is a dict: {"intent": UserUttered.intent,
#                   "entities": UserUttered.entities,
#                   "text": text}
self.latest_message = latest_message if latest_message else {}
```

Ghostvv:
@smn-snkl Hi, do you plan to continue working on this PR?

smn-snkl:
@Ghostvv, yes generally I am. I’m a bit busy for a few more days. Would look into it then unless you say it’s very urgent?

Ghostvv:
@smn-snkl thanks, for the update. It doesn't seem to be urgent right now.

tmbo:
@smn-snkl we'll need to get this merged till next Tuesday as we will restructure the code afterwards and the code will go stale. I think the best thing would be to merge it before then. But if you don't have enough time to complete it, we'll need to close this PR and redoing the change will probably cost more time.

PR #1683, opened 2019-02-05
Updated migration docs with Rasa Core 0.13.0

### LINKED ISSUE #1677: Migration docs for rasa_core 0.12.x - > 0.13
Detail what changes are necessary to upgrade

MetcalfeTom:
To fix #1677


docs/migrations.rst
```diff
@@ -7,6 +7,27 @@ Migration Guide
...
 
+.. _migration-to-0-13-0:
+
+0.12.x to 0.13.0
+----------------
+
+.. warning::
+
+    This is major new version with a lot of changes under the hood as well
+    as on the API level. Please take a careful look at the mentioned
```
akelad: hmm, i don't think `with a lot of changes under the hood as well as on the API level` is true 😄 
I think the warning should say we dropped support for py2 and to retrain their models with py3 when switching

akelad:
also i believe this should be based off of 0.13.x

MetcalfeTom:
> also i believe this should be based off of 0.13.x

Is there any downside to merging this, then starting a new minor from `master`?  If I PR to `0.13.x`, I'm going to be merging twice.

akelad:
we don't release minors from master, as it may contain breaking changes. You just merge the 0.13.x branch into master afterwards (and pray for no merge conflicts, otherwise create a new branch off of 0.13.x, merge master into it and then merge it into master)

akelad:
so it's approved now, but please merge it into 0.13.x rather than into master.

MetcalfeTom:
(Merged for version `rasa_core==0.13.2`)

PR #1685, opened 2019-02-05
Added migration text for v0.13

MetcalfeTom:


akelad:
@MetcalfeTom do you have any idea why the PR build is breaking?

MetcalfeTom:
There was an extra travis step in the `master` branch which wasn't removed when I rebased, opening another PR now

PR #1695, opened 2019-02-06
Merge 0.13.2 into master

MetcalfeTom:


MetcalfeTom:
Good idea, we have more things to remove from the docs now that we've dropped py2

Issue #1596, opened 2019-01-18
Create a policy that maps an intent to an action


akelad:
The idea is that you'll be able to provide a dictionary of intent/action maps, so that every time one of these intents is identified Rasa Core will directly respond with that action regardless of the context

Ghostvv:
doesn't memo with `max_history=1` is exactly this?

Ghostvv:
it is a bit more than that

akelad:
exactly :P

imLew:
made PR for this #1701

PR #1701, opened 2019-02-07
Mapping policy

imLew:
**Proposed changes**:
- Add the `MappingPolicy` which allows the direct mapping of intents to actions by using the new `maps_to` keyword for intents in the domain file. Mapped are prioritized over all other actions. After the mapped actions has been executed the agent returns to `action_listen`.

fixes https://github.com/RasaHQ/rasa_core/issues/1596

- [x] updated the changelog


imLew:
> ok i just noticed you had requested a review, but didn't actually tick that it's ready for review... so is it ready for review or not?
yes, i is, only forgot to update 

> i'm submitting the comments I started making already, but in general the default action we discussed is still missing and so are the docs etc
what should the default action be?

akelad:
@imLew it should be the ActionBack thing we discussed. So basically reverting two user utterances if the user uses the intent e.g. `/back`. So this should be a default action, like `action_restart` and then we can have an example in the docs about how you can use that action to go back

docs/migrations.rst
```diff
@@ -9,6 +9,16 @@ how you can migrate from one version to another.
 
 .. _migration-to-0-13-0:
 
+0.13.x to 0.14.0
+
+General
+~~~~~~~
+
+-The `MappingPolicy` is now included in `default_config.yml`. If you are using
```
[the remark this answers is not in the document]
tmbo: yes it did not exist before, but its functionality did (e.g. mapping /restart to restart action)

rasa_core/constants.py
```diff
@@ -22,6 +22,8 @@
 
 FORM_SCORE = 1.3
 
+MAPPING_SCORE = 1.4
```
tmbo: needs to be merged with https://github.com/RasaHQ/rasa_core/pull/1735 (I guess whoever merges first into master wins :D )

rasa_core/domain.py
```diff
@@ -196,6 +196,9 @@ def collect_intent_properties(intent_list):
         intent_properties = {}
         for intent in intent_list:
             if isinstance(intent, dict):
+                if 'use_entities' not in list(intent.values())[0].keys():
```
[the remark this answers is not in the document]
tmbo: I guess it wouldn't hurt to do
```python
for properties in intent.values():
  if "use_entities" not in properties:
    properties["use_entities"] = true

intent_properties.update(intent)
```

rasa_core/policies/mapping_policy.py
```diff
@@ -0,0 +1,66 @@
...
+from rasa_core.trackers import DialogueStateTracker
+from rasa_core.constants import MAPPING_SCORE
+
+logger = logging.getLogger(__name__)
+
+
+class MappingPolicy(Policy):
+    """Policy which maps intents directly to actions.
+
+    Intents can be assigned actioins in the domain file which are to be
```
tmbo:
```suggestion
    Intents can be assigned actions in the domain file which are to be
```

rasa_core/policies/mapping_policy.py
```diff
@@ -0,0 +1,66 @@
...
+                                     tracker: DialogueStateTracker,
+                                     domain: Domain) -> List[float]:
+        """Predicts the assigned action.
+
+        If the current intent is assigned to an action that action will be
+        predicted with the highest probability of all policies. If it is not
+        the policy will predict zero for every action."""
+
+        intent = tracker.latest_message.intent.get('name')
+        action_name = domain.intent_properties.get(intent, {}).get('maps_to')
```
tmbo: @amn41 any native speaker insights if this is the best property name to signal that an action is always going to follow this intent?
```
intents:
  hello:
    maps_to: action_greet
```

rasa_core/policies/mapping_policy.py
```diff
@@ -0,0 +1,66 @@
...
+
+        prediction = [0.0] * domain.num_actions
+        if action_name is not None:
+            if self.last_action_name == action_name:
+                idx = domain.index_for_action(ACTION_LISTEN_NAME)
+                prediction[idx] = MAPPING_SCORE
+            else:
+                idx = domain.index_for_action(action_name)
```
tmbo: we need to validate this, it is user input: probably best to do this when reading the domain

rasa_core/policies/mapping_policy.py
```diff
@@ -0,0 +1,66 @@
...
+from rasa_core import utils
+from rasa_core.domain import Domain
+from rasa_core.policies.policy import Policy
+from rasa_core.trackers import DialogueStateTracker
+from rasa_core.constants import MAPPING_SCORE
+
+logger = logging.getLogger(__name__)
+
+
+class MappingPolicy(Policy):
```
tmbo: should this by default map `restart` intent to `ActionRestart`? I'd love to get this handled properly as part of this

rasa_core/policies/mapping_policy.py
amn41: yeah sounds fine to me! other option would be `triggers`

docs/migrations.rst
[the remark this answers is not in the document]
akelad: does that require it going in the migrations guide then though?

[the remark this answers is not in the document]
tmbo: depends, if it does handle the restart than the restart would stop to work if you specified a pipeline in a config. you'd need to manually add it there.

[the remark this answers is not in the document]
akelad: ahhh ok, got it now

examples/moodbot/nlu_model_config.yml
```diff
@@ -1 +1,3 @@
+langue: "en"
```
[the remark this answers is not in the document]
imLew:
```suggestion
```
nope, just an artefact

rasa_core/policies/mapping_policy.py
imLew: Will change it to `triggers` as that seems more universal, `maps_to` is math-speak

rasa_core/policies/mapping_policy.py
imLew: what do we need to validate?
`triggers` mapping to a valid action?

tmbo: yes, it is user input so the action name might e.g. be misspelled, idx will be none in this case which needs to be handled. In addition to that, we should make sure the value for the trigger is a proper action when we read the domain file.

imLew: but the name could only be mispelled in the domain file, so checking that it is correct and indeed a valid action both happen at the same time right?

tmbo: yes exactly - and we should inform the user right when that happens.

nevertheless, I'd like to handle the `None` value here as well in some way (because at some point for some reason we will get a None and then the exception raised will be odd)

rasa_core/constants.py
erohmensing: yeah I'm about to merge mine but @imLew I can help you adapt yours to fit mine, should just require adding a default priority of 5 to this policy's` __init__` i think, and then using `1.0` instead of `MAPPING_SCORE`

imLew:
> @imLew it should be the ActionBack thing we discussed. So basically reverting two user utterances if the user uses the intent e.g. `/back`. So this should be a default action, like `action_restart` and then we can have an example in the docs about how you can use that action to go back

@akelad Can we discuss this on Monday? I'm looking at the `trackers` right now but don't know which variables are relevant here and basically have no clue what the canonical way of implementing this action would be.

tmbo:
@imLew to revert user messages we use the events on the tracker. The tracker keeps all modification of the state in its `tracker.events`. From this list of events (which is append only, we never change existing events but always only append new ones) we then calculate the current state of the conversation (e.g. what slot values are set, what was the last user message or action). To "revert" a user message we append the event `UserUtteranceReverted` event to the trackers events. When we calculate the state, it will look like as if the user never wrote that message and we are waiting for the users input again.

In the action_back, you'd need to return that event in the run method.

Ghostvv:
I think in `action_back`, you'd need to return that event `UserUtteranceReverted` 2 times

docs/migrations.rst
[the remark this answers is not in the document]
imLew: since it does now handle '/restart' i guess we will leave it in here

tmbo:
Please request another review once your good to go

rasa_core/policies/mapping_policy.py
```diff
@@ -0,0 +1,69 @@
...
+    """Policy which maps intents directly to actions.
+
+    Intents can be assigned actions in the domain file which are to be
+    executed whenever the intent is detected. This policy takes precedence over
+    any other policy."""
+
+    def __init__(self, priority: Optional[int] = 5) -> None:
```
erohmensing:
```suggestion
    def __init__(self, priority: int = 5) -> None:
```
At least all of the other specified policies (not policy class) look like this

rasa_core/actions/action.py
```diff
@@ -157,6 +159,18 @@ def __str__(self) -> Text:
...
 
 
+class ActionBack(Action):
+    """Revert the tracker state by two user utterances."""
+
+    def name(self) -> Text:
+        return ACTION_BACK_NAME
+
+    def run(self, dispatcher, tracker, domain):
+        from rasa_core.events import UserUtteranceReverted
```
akelad: don't need to add this import statement, it already gets imported at the start

rasa_core/domain.py
```diff
@@ -196,9 +229,13 @@ def collect_intent_properties(intent_list):
...
         for intent in intent_list:
             if isinstance(intent, dict):
+                for properties in intent.values():
+                    if 'use_entities' not in properties:
+                        properties['use_entities'] = True
                 intent_properties.update(intent)
             else:
-                intent_properties.update({intent: {'use_entities': True}})
+                intent = {intent: {'use_entities': True}}
+                intent_properties.update(intent)
```
akelad: any reason this isn't just what it was originally? `intent_properties.update({intent: {'use_entities': True}})`

rasa_core/actions/action.py
```diff
@@ -157,6 +159,18 @@ def __str__(self) -> Text:
...
+
+        return [UserUtteranceReverted(), UserUtteranceReverted()]
```
akelad: ok I know i said this would be fine but i just tested it out and i feel like we need some sort of utterance or something that tells the user they've gone back. Or repeat the previous action? not 100% sure what the best UX here is -- @Ghostvv thoughts?

rasa_core/policies/mapping_policy.py
```diff
@@ -0,0 +1,69 @@
...
+            intent = tracker.latest_message.intent.get('name')
+            action = domain.intent_properties.get(intent, {}).get('triggers')
+            if action:
+                idx = domain.index_for_action(action)
+                if idx is None:
+                    logger.warning("MappingPolicy tried to predict unkown "
+                                   "action '{}'.".format(action))
+                else:
+                    prediction[idx] = 1
+            elif tracker.latest_message.intent.get('name') == 'restart':
```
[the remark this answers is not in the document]
tmbo: you mean adding `restart` to the intent list?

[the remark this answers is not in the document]
imLew: @akelad we thought about it but nothing was decided yet
@tmbo yes, `restart` should be an intent and so should `back`, i think this makes sense because:
- they are entered the same as other inputs, which means they should be handled the same
- we might want to add more of these utility/default intents in the future (like resetting slots for example)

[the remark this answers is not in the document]
akelad: yes adding `restart`(along with its mapping) to the intent list in the domain file. i was sure that's what we'd decided on :P rather than explicitly comparing for it here

rasa_core/actions/action.py
imLew: We should not just replay the last action, I think that would be pretty confusing if it doesn't come with a message.
If verbosity is not a problem we could do an utterance 
"OK, I forgot the last thing you said, just like you asked.
Now, where were we?
You said: [2nd to last user input before /back]
and I said [2nd to last action]"

rasa_core/policies/mapping_policy.py
```diff
@@ -0,0 +1,69 @@
...
+                idx = domain.index_for_action(ACTION_RESTART_NAME)
+                prediction[idx] = 1
+            elif tracker.latest_message.intent.get('name') == 'back':
```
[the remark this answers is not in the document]
imLew: Should `restart` be a default though?
We could just include the following in the docs
`intents:
   - back: {triggers: action_back}
   - restart {triggers: action_restart}`

docs/policies.rst
```diff
@@ -196,6 +196,19 @@ training data. It predicts the next action with confidence ``1.0``
...
+The ``MappingPolicy`` can be used to directly map intents to actions such that
+the mapped action will always be executed. The mappins are assigned by giving
+and intent the property `'triggers'`, e.g.:
+
+.. code-block:: yaml
+
+  intents:
+   - greet: {triggers: utter_goodbye}
+
+An intent can only be mapped to at most one action.
```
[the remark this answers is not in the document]
erohmensing: Human: Hello
Bot: Bye

I know it's just a doc example, but surely `greet` should trigger `utter_greet`, no? 😄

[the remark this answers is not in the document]
imLew: We could change the policy to execute ActionBack after it executed a mapped action. 
But that might mes up the tracker if the mapping is in the examples right? So we have to somehow make this clear in the docs.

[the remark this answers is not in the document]
imLew: in the latest commit I did just that, executing action_back after if the last action is the action that the current intent in mapped to

rasa_core/policies/mapping_policy.py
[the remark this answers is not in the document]
akelad: restart needs to 100% be a default, it always has been and should continue to be :P

docs/policies.rst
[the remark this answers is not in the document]
akelad: no no no i don't think the policy should execute ActionBack after it executed a mapped action. I'm sure we discussed this, that's going to cause some undesired behaviour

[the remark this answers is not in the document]
akelad: We should explain the `UserUtteranceReverted` thing in the docs though

rasa_core/actions/action.py
akelad: I'm not sure how much we should go into this though, we shouldn't overcomplicate this. @Ghostvv any thoughts?

Ghostvv: these two `UserUtteranceReverted` will not cause to repeat previous action

Ghostvv: I suggest the same as with restart, utilize `utter_back` template:
```
        # only utter the template if it is available
        dispatcher.utter_template("utter_back", tracker,
                                  silent_fail=True)
```

akelad: I know it's not going to cause a repetition of the previous action, I was asking whether we should do that :P

akelad: also we decided on `/back` not being mapped to `action_back` by default right?

Ghostvv: no, by default I don't think we should repeat previous action, I think utterance should be enough

We could add to the docs example, that if you want to repeat previous action, return `[UserUtteranceReverted(), UserUtteranceReverted(), ActionReverted()]`

rasa_core/policies/mapping_policy.py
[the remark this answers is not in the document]
Ghostvv: I agree, we keep `ActionBack` but don't add default intent for it

rasa_core/actions/action.py
akelad: ok cool

rasa_core/domain.py
imLew: Yes, I'm not 100% sure if it's necessary but the reason was that this function modifies the intents. Previously those intents that did not have properties added in the domain would not have properties after this function was called either, even though they are supposed to have the property `use_entities`.
It might not matter though because properties of intents are afaik always accessed through the `domain.intent_properties` dictionary.

rasa_core/actions/action.py
imLew: i removed the import in the latest commit

docs/policies.rst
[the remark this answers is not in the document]
imLew: Would it not be better to have the policy append `UserUtteranceReverted`? Otherwise mapped actions could not be normal actions and vice versa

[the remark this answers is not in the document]
imLew: Actually that wouldn't work because the action is executed only after any policy methods run.
But then how can we make the action return `UserUtteranceReverted` if it was performed based on the mapping policy?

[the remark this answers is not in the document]
akelad: @imLew this is up to the user to implement in their own action.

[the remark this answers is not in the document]
imLew: That would mean that if users want to map to an already existing action they would have to create a copy that also returns UserUtteranceReverted.
If it always needs to be done anyway I think we should have the policy do it, maybe by just having an action that returns only one `UserUtteranceRevert`.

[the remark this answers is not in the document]
akelad: I'm not sure I agree, there may be cases where people want to change that behaviour. @Ghostvv what are your thoughts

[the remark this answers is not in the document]
tmbo: I don't see why we need this now. I think we should not implement any revert behaviour for all mapped actions. This will be working just fine for the use cases we need it for atm (e.g. restart and back) and we can change this once we discover that it really is not working for other actions.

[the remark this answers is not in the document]
Ghostvv: I can see both arguments. I agree it is not a necessity. I guess if we want it, we could create action wrapper in the mapping policy, that when passed mapping actions, creates a new action with appended `UserUtteranceReverted()`. The problem with that, is all events returned by original action will be forgotten, so no slots would be set. This wrapper could insert  `UserUtteranceReverted()` in the beginning though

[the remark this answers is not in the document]
akelad: Yeah let's not implement the revert behaviour for now, but just add a note to the docs about it

rasa_core/actions/action.py
akelad: @tmbo @Ghostvv do we want `/back` to be a default mapping or no?

Ghostvv: I don't have strong open opinion here, why not?

akelad: idk, we decided at some point we didn't want to do that for some reason. But let's just add it anyways

tmbo:
we should get this merged wednesday @imLew

rasa_core/actions/action.py
Ghostvv: if we already have an action, why not to add intent

erohmensing:
@imLew next time you merge master, please move mapping policy to after formpolicy (added today) instead of after memo in the policies docs 😄

Issue #1061, opened 2018-09-24
Add support for PostgreSQL tracker store


stevus:

akelad:
Any more details on this? We're happy to accept a PR for this, but probably won't be building it ourselves for now

stevus:
Yea I'd be happy to jump on this one.

PR #1717, opened 2019-02-12
SQL Tracker Store

### LINKED ISSUE #1713: Create an SQL tracker store

MetcalfeTom:
**Proposed changes**:
- fixes #1713 
- fixes https://github.com/RasaHQ/rasa_core/issues/1061

- [x] updated the changelog


tmbo:
@MetcalfeTom how is the PR coming along?

MetcalfeTom:
Gotta update the `test_dialogues` to get the tests to run, since the events in there don't have the necessary attributes that my tracker store looks for (e.g. `timestamp`) - I will open a separate PR for that.

I'm hoping to automate it by using a script to create a tracker from the dialogues, then dumping them into JSON format.  Is there a simpler way?

tmbo:
please let me know when this is ready for another review (by requesting another review)

docs/tracker_stores.rst
```diff
@@ -94,7 +131,7 @@ MongoTrackerStore
 
                 python -m rasa_core.run --core models/dialogue --endpoints endpoints.yml
 :Parameters:
-    - ``url`` (default: ``mongodb://localhost:27017``): URL of your MongoDB
+    - ``host`` (default: ``mongodb://localhost:27017``): URL of your MongoDB
```
tmbo: You can't just change these parameters for mongo & redis (you didn't change the code, and I also don't think we should).

rasa_core/events/__init__.py
```diff
@@ -204,8 +204,12 @@ def __init__(self, text,
 
     @staticmethod
     def _from_parse_data(text, parse_data, timestamp=None, input_channel=None):
-        return UserUttered(text, parse_data["intent"], parse_data["entities"],
-                           parse_data, timestamp, input_channel)
+        return UserUttered(text,
+                           parse_data.get("intent"),
+                           parse_data.get("entities", {}),
```
tmbo:
```suggestion
                           parse_data.get("entities", []),
```

rasa_core/tracker_store.py
```diff
@@ -39,6 +40,11 @@ def find_tracker_store(domain, store=None, event_broker=None):
                                      host=store.url,
                                      event_broker=event_broker,
                                      **store.kwargs)
+        elif store.type == 'SQL':
```
tmbo:
```suggestion
        elif store.type.lower() == 'sql':
```

rasa_core/tracker_store.py
```diff
@@ -39,6 +40,11 @@ def find_tracker_store(domain, store=None, event_broker=None):
...
+            return SQLTrackerStore(domain=domain,
+                                   host=store.url,
```
tmbo: well either use url or host, according to docs the user would supply `store.host`

docs/tracker_stores.rst
```diff
@@ -26,6 +26,43 @@ InMemoryTrackerStore (default)
...
+
+    3. To start the Rasa Core server using your SQL backend,
+       add the :code:`--endpoints` flag, e.g.:
+
+        .. code-block:: bash
+
+            python -m rasa_core.run --core models/dialogue --endpoints endpoints.yml
+:Parameters:
+    - ``dialect`` (default: ``sqlite``): The dialect used to communicate with your SQL backend
+    - ``host`` (default: ``None``): URL of your SQL database
```
tmbo: let's use url and not host

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+
+
+class SQLTrackerStore(TrackerStore):
+    """Store which can save and retrieve trackers from an SQL database"""
+
+    from sqlalchemy.ext.declarative import declarative_base
+    Base = declarative_base()
+
+    class SQLEvent(Base):
+        __tablename__ = 'conversations'
```
tmbo: should this rather be `events`?

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+        timestamp = Column(Float)
+        intent_name = Column(String)
+        action_name = Column(String)
+        data = Column(String)
+
+    def __init__(self,
+                 domain: Optional[Domain] = None,
+                 dialect: Text = 'sqlite',
+                 host: Text = None,
+                 event_broker: Optional[EventChannel] = None,
```
tmbo: let's use same param order for all stores, so probably this should be the last argument

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+                 db: Text = 'rasa.db',
+                 username: Text = None,
+                 password: Text = None) -> None:
+        from sqlalchemy.orm import sessionmaker
+        from sqlalchemy.engine.url import URL
+        from sqlalchemy import create_engine
+
+        engine_url = URL(dialect, username, password, host, database=db)
```
tmbo: what if the username & password are part of the host string, will this work?

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+
+        logger.debug('Attempting to connect to database '
+                     'via "{}"'.format(engine_url.__to_string__()))
+
+        self.engine = create_engine(engine_url)
+        self.session = sessionmaker(bind=self.engine)()
+
+        logger.debug('Connection successful, ensuring conversations table...')
```
tmbo:
```suggestion
        logger.debug('Connection to SQL db successful, ensuring conversations table exists...')
```

tmbo: this logging stmt should be moved into the `_ensure_table` function

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+
+        self._ensure_table()
+        super(SQLTrackerStore, self).__init__(domain, event_broker)
+
+        logger.debug('SQL tracker store successfully initialised')
```
tmbo:
```suggestion
```

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+
+    def _ensure_table(self) -> None:
+        """Creates the conversations table in the database if not present"""
+
+        metadata = MetaData()
+
+        Table("conversations", metadata,
```
tmbo: do we need these field definitions twice? I mean they are already part of the `SQLEvent ` class

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+              Column("sender_id", String, nullable=False),
+              Column("type_name", String, nullable=False),
+              Column("timestamp", Float),
+              Column("intent_name", String),
+              Column("action_name", String),
+              Column("data", String))
+
+        metadata.create_all(self.engine)
+
+    def keys(self) -> List:
```
tmbo:
```suggestion
    def keys(self) -> List[Text]:
```

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+        """Returns the keys of the items stored in the database"""
+        return self.SQLEvent.__table__.columns.keys()
+
+    def retrieve(self, sender_id: Text) -> DialogueStateTracker:
+        """Creates a tracker from all previously stored events"""
+
+        subquery = self.session.query(self.SQLEvent)
+        query = subquery.filter_by(sender_id=sender_id).all()
```
tmbo: this is not the query, it is the result isn't it?

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+        events = [json.loads(event.data) for event in query]
+
+        if self.domain and len(events) > 0:
+            logger.debug('Recreating tracker '
```
tmbo: this isn't a useful debug message, should either include the sender id or be removed

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+                         'from {} stored events'.format(len(events)))
+
+            tracker = DialogueStateTracker.from_dict(sender_id, events,
```
tmbo:
```suggestion
            return DialogueStateTracker.from_dict(sender_id, events,
```

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+                                                     self.domain.slots)
+        else:
+            logger.warning("Can't retrieve tracker from SQL storage.  "
+                           "Returning `None` instead.")
+            tracker = None
```
tmbo:
```suggestion
            return None
```

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+        return tracker
```
tmbo:
```suggestion
```

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+        events = [json.loads(event.data) for event in query]
+
+        if self.domain and len(events) > 0:
+            logger.debug('Recreating tracker '
+                         'from {} stored events'.format(len(events)))
+
+            tracker = DialogueStateTracker.from_dict(sender_id, events,
+                                                     self.domain.slots)
+        else:
+            logger.warning("Can't retrieve tracker from SQL storage.  "
```
tmbo: same, should include sender id

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+                           "Returning `None` instead.")
+            tracker = None
+        return tracker
+
+    def save(self, tracker: DialogueStateTracker) -> None:
+        """Updates database with events from the current conversation"""
```
tmbo:
```suggestion
        """Update database with events from the current conversation."""
```

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+              Column("data", String))
+
+        metadata.create_all(self.engine)
+
+    def keys(self) -> List:
+        """Returns the keys of the items stored in the database"""
+        return self.SQLEvent.__table__.columns.keys()
+
+    def retrieve(self, sender_id: Text) -> DialogueStateTracker:
+        """Creates a tracker from all previously stored events"""
```
tmbo:
```suggestion
        """Create a tracker from all previously stored events."""
```

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+              Column("type_name", String, nullable=False),
+              Column("timestamp", Float),
+              Column("intent_name", String),
+              Column("action_name", String),
+              Column("data", String))
+
+        metadata.create_all(self.engine)
+
+    def keys(self) -> List:
+        """Returns the keys of the items stored in the database"""
```
tmbo:
```suggestion
        """Collect all keys of the items stored in the database."""
```

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+
+        logger.debug('Connection successful, ensuring conversations table...')
+
+        self._ensure_table()
+        super(SQLTrackerStore, self).__init__(domain, event_broker)
+
+        logger.debug('SQL tracker store successfully initialised')
+
+    def _ensure_table(self) -> None:
+        """Creates the conversations table in the database if not present"""
```
tmbo:
```suggestion
        """Create the conversations table in the database if not present."""
```

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
 
     def keys(self):
         return [c["sender_id"] for c in self.conversations.find()]
+
+
+class SQLTrackerStore(TrackerStore):
+    """Store which can save and retrieve trackers from an SQL database"""
```
tmbo:
```suggestion
    """Store which can save and retrieve trackers from an SQL database."""
```

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+            tracker = None
+        return tracker
+
+    def save(self, tracker: DialogueStateTracker) -> None:
+        """Updates database with events from the current conversation"""
+
+        if self.event_broker:
+            self.stream_events(tracker)
+
+        events = self._event_buffer(tracker)  # only store recent events
```
tmbo: please find a better name, e.g. `addtional_events` or something

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+
+        for event in events:
+            try:
+                data = event.as_dict()
+            except AttributeError as e:
```
tmbo: when does this happen? I don't think we do this anywhere else

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+            timestamp = data.get("timestamp")
+
+            self.session.add(self.SQLEvent(sender_id=tracker.sender_id,
+                                           type_name=event.type_name,
+                                           timestamp=timestamp,
+                                           intent_name=intent,
+                                           action_name=action,
+                                           data=json.dumps(data)))
+        self.session.commit()
+        logger.debug('Tracker stored to database')
```
tmbo:
```suggestion
```

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+
+        latest_events = []
+
+        for event in reversed(tracker.events):
+            if event.timestamp > max_timestamp:
+                latest_events.append(event)
+            else:
+                break
+
+        logger.debug('Storing {} recent events'.format(len(latest_events)))
```
tmbo:
```suggestion
```

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,137 @@ def retrieve(self, sender_id):
...
+        from sqlalchemy import func
+        query = self.session.query(func.max(self.SQLEvent.timestamp))
+        max_timestamp = query.filter_by(sender_id=tracker.sender_id).scalar()
+
+        if max_timestamp is None:
+            max_timestamp = 0
+
+        latest_events = []
+
+        for event in reversed(tracker.events):
```
tmbo: why do we reverse this twice, can we not just remove both `reversed`? (here and L392)

setup.py
```diff
@@ -67,7 +67,8 @@
     "python-socketio~=3.0",
     "pydot~=1.4",
     "keras-applications==1.0.6",
-    "keras-preprocessing==1.0.5"
+    "keras-preprocessing==1.0.5",
+    "sqlalchemy~=1.2.17"
```
tmbo:
```suggestion
    "sqlalchemy~=1.2"
```

tests/test_trackers.py
```diff
@@ -29,13 +32,17 @@ def __init__(self, domain):
 
 
 def stores_to_be_tested():
-    return [MockRedisTrackerStore(domain),
-            InMemoryTrackerStore(domain)]
+    with tempfile.TemporaryDirectory() as temp:
+        return [MockRedisTrackerStore(domain),
+                InMemoryTrackerStore(domain),
+                SQLTrackerStore(domain,
+                                db='{}rasa.db'.format(temp))]
```
tmbo:
```suggestion
                                db=os.path.join(temp, 'rasa.db'))]
```

docs/tracker_stores.rst
MetcalfeTom: I'm confused, in the code for mongo & redis, the parameter is `host`.  How does `url` get parsed as `host` when we include it in the `endpoints.yml`?

rasa_core/tracker_store.py
MetcalfeTom: Yeah I went back and forth about this...although `events` is probably better since when this is in production, the db will have multiple conversations stored insequentially

rasa_core/tracker_store.py
MetcalfeTom: OK, but I'm not sure about putting "exists" - if the table doesn't exist, then `_ensure_table` will create it

rasa_core/tracker_store.py
MetcalfeTom: Are you OK with me just completely removing this `return` statement then?  This line would have no effect

rasa_core/tracker_store.py
MetcalfeTom: Happened during testing, but it has been fixed since `test_dialogues` were updated, I'll remove it now

rasa_core/tracker_store.py
MetcalfeTom: 1. We assume that most of the tracker is already in the store, in which case it's quicker to go from back to the front
2. If we don't reverse it here, we have to do a list comprehension to return the latest events once we reach the timestamp, i.e. iterate over `tracker.events` twice

rasa_core/tracker_store.py
MetcalfeTom: Why remove this _and_ the other saving debug statement?  Then there are no logs for this process

docs/tracker_stores.rst
```diff
@@ -26,6 +26,43 @@ InMemoryTrackerStore (default)
 :Configuration:
     To use the `InMemoryTrackerStore` no configuration is needed.
 
+SQLTrackerStore
+~~~~~~~~~~~~~~~~~~
+
+:Description:
+    `SQLTrackerStore` can be used to store the conversation history in an SQL database.
```
wochinge:
```suggestion
    ``SQLTrackerStore`` can be used to store the conversation history in an SQL database.
```

docs/tracker_stores.rst
```diff
@@ -26,6 +26,43 @@ InMemoryTrackerStore (default)
 :Configuration:
     To use the `InMemoryTrackerStore` no configuration is needed.
 
+SQLTrackerStore
+~~~~~~~~~~~~~~~~~~
```
wochinge:
```suggestion
~~~~~~~~~~~~~
```

docs/tracker_stores.rst
```diff
@@ -26,6 +26,43 @@ InMemoryTrackerStore (default)
...
+:Configuration:
+    To set up Rasa Core with SQL the following steps are required:
+
+    1. Add required configuration to your `endpoints.yml`
+
+        .. code-block:: yaml
+
+            tracker_store:
+                type: SQL
+                dialect: <the dialect used to interact with the db, e.g. `sqlite`.  [Consult the SQLAlchemy docs for available dialects](https://docs.sqlalchemy.org/en/latest/core/engines.html#database-urls)>
```
wochinge: does the link actually work within a code block?

wochinge: rather add the link in your Parameters thing down below

rasa_core/tracker_store.py
```diff
@@ -250,3 +256,136 @@ def retrieve(self, sender_id):
...
+class SQLTrackerStore(TrackerStore):
+    """Store which can save and retrieve trackers from an SQL database."""
+
+    from sqlalchemy.ext.declarative import declarative_base
+    Base = declarative_base()
+
+    class SQLEvent(Base):
+        __tablename__ = 'conversations'
+
+        id = Column(Integer, primary_key=True)
```
wochinge: why not sender_id as primary key?

rasa_core/tracker_store.py
MetcalfeTom: Managed to remove the `_ensure_table` method

rasa_core/tracker_store.py
MetcalfeTom: As in, the host string looks like: `user:pass@mssql2017:1433`?  Yes that will work.

setup.py
```diff
@@ -68,8 +68,9 @@
     "pydot~=1.4",
     "keras-applications==1.0.6",
     "keras-preprocessing==1.0.5",
+    "sqlalchemy~=1.2",
     "kafka-python~=1.4",
-    "sklearn-crfsuite~=0.3.6",
+    "sklearn-crfsuite~=0.3.6"
```
tmbo:
```suggestion
    "sklearn-crfsuite~=0.3.6",
```

PR #1727, opened 2019-02-16
Resolve #1720

### LINKED ISSUE #1720: Slot name with "illegal character" should be flagged?
**Rasa Core version**:
0.12.1 (but the master has the same issue)

**Python version**: 
3.6.6

Linux

**Issue**:
I had slot name with dot e.g. `my_company_name.special_slot`
this works as long as it is not used in a template but because the substitute is done with python built-in `format`, dot notation won't work.

https://github.com/RasaHQ/rasa_core/blob/master/rasa_core/nlg/template.py#L79

I did the easy workaround of removing dot but perhaps either 
(a) allow dot in slot by updating `_fill_template_text` function or 
(b) when domain file is passed, check and see if slot name with invalid python variable name is also used in template. if so error out
Would be helpful

naoko:
Handles slot name contains character that is invalid as python variable name (e.g. dot) in template

**Proposed changes**:
- fixes #1720 

- [x] updated the changelog


rasa_core/nlg/template.py
```diff
@@ -76,7 +77,10 @@ def _fill_template_text(
         # Filling the template variables in the template
         if template_vars:
             try:
-                template["text"] = template["text"].format(**template_vars)
+                # blacklist characters that probably should not be
+                # part of slot name and replace with old %-formatting
+                text = re.sub(r'{([^\n,]+?)}', r'{0[\1]}', template["text"])
```
wochinge: @naoko I thought we agreed on not blacklisting the characters but instead handling it as it is done here https://stackoverflow.com/questions/7934620/python-dots-in-the-name-of-variable-in-a-format-string#comment9695339_7934969 ?

tests/test_nlg.py
```diff
@@ -79,3 +81,45 @@ def test_nlg_schema_validation_empty_buttons():
...
+@pytest.mark.parametrize("filled_slots", [
+    {"tag_w_underscore": "a"},
+    {"tag with space": "bacon", "tag_w_int_val": 123},
+    {"tag.with.dot": "chocolate", "tag_2": "b"},
+    {"tag-w-dash": "apple pie", "tag.with.float.val": 1.3},
+    {"tag-w-$": "banana"},
+    {"tag-w-@": "one", "tagCamelCase": "two", "tag-w-*": "three"},
+])
+def test_nlg_fill_template_text(filled_slots):
+    template_text = ", ".join(["{" + t + "}" for t in filled_slots.keys()])
```
wochinge: You can also use parametrize a bit differently so you don't have to do the strings joins (https://docs.pytest.org/en/latest/parametrize.html):

```python
@pytest.mark.parametrize("slot_name,slot_value", [("tag_w_underscore", "a"),]
def test_nlg_fill_template_text(slot_name, slot_value): 
   pass 
```

tests/test_nlg.py
```diff
@@ -79,3 +81,45 @@ def test_nlg_schema_validation_empty_buttons():
...
+    resolved_text = ", ".join([str(s) for s in filled_slots.values()])
+
+    template = {'text': template_text}
+    t = TemplatedNaturalLanguageGenerator(templates=dict(
```
wochinge: you can also leave the constructor empty for the tests

tests/test_nlg.py
```diff
@@ -79,3 +81,45 @@ def test_nlg_schema_validation_empty_buttons():
...
+    template_text = ", ".join(["{" + t + "}" for t in filled_slots.keys()])
+    template = {'text': template_text}
+    t = TemplatedNaturalLanguageGenerator(templates=dict(
+        bot_message_1=[template]
+    ))
+    result = t._fill_template_text(
+        template=template,
+        filled_slots=filled_slots
+    )
+    assert mock_logger.exception.call_count == 1
```
wochinge: you can also skip the mocking and simply assert that the response is the same as the template string

rasa_core/nlg/template.py
naoko: My attempt was to yes handling showing in stackoverflow. this regex sub is to change `{tag}` to `{0[tag]}`. the only blacklisting I'm doing here is to make sure not to pick up "tag name" with newline and "," as I guess this could be bad assumption but I thought tag name (slot name with new line or , ) would make no sense.

CHANGELOG.rst
```diff
@@ -30,6 +30,7 @@ Removed
 
 Fixed
 -----
+- Handles slot name contains characters that are invalid as python variable name (e.g. dot) in a template
```
```suggestion
- Handles slot names which contain characters that are invalid as python variable name (e.g. dot) in a template
```

tests/test_nlg.py
```diff
@@ -79,3 +80,53 @@ def test_nlg_schema_validation_empty_buttons():
 def test_nlg_schema_validation_empty_image():
     content = {"text": "Hey there!", "image": None}
     assert CallbackNaturalLanguageGenerator.validate_response(content)
+
+
+@pytest.mark.parametrize("filled_slots", [
```
wochinge: @naoko Please also change this one to `pytest.mark.parametrize("slot_name, slot_value"`

tests/test_nlg.py
```diff
@@ -79,3 +80,52 @@ def test_nlg_schema_validation_empty_buttons():
...
+    assert result == {'text': str(slot_value)}
+
+
+@pytest.mark.parametrize("slot_name, slot_value", [
+    ("tag_w_underscore", "a"),
+    ("tag.with.float.val", 1.3),
+    ("tag-w-$", "banana"),
+    ("tagCamelCase", "two"),
+])
+def test_nlg_fill_template_text2(slot_name, slot_value):
```
wochinge: Please merge this with `test_nlg_fill_template_text2`

PR #1735, opened 2019-02-19
Add policy priority property

erohmensing:
**Proposed changes**:
- Add `priority`, a property of `Policy`, which is used in the case of equal max confidence (i.e. when memo, fallback, and form policies all have max confidences of 1.0)
- Add default priorities for all of our supported policies
- Make policy priority configurable, but throw a warning if multiple policies in an ensemble have the same priority

- [x] updated the changelog

Documentation will be updated in a full update of the `Training and Policies` page since it has more problems than just introducing this.


erohmensing:
@akelad something I definitely want your feedback on:
- default values for our policies -- I figured that keeping them all different would be good, albeit somewhat arbitrary. but maybe, for example, `FallbackPolicy` and `TwoStageFallbackPolicy` should be the same since you cant use them together? Are there any other policy combinations that can't be used together?

akelad:
@erohmensing i'd say Keras and Embedding and Sklearn can also get the same priority

erohmensing:
Okay, that's what I originally had but wasn't sure if anyone would attempt to use them together for some strange reason

akelad:
Memoization and AugmentedMemo also the same

erohmensing:
Yeah, AugmentedMemo will have the same as Memo already since it doesn't have its own __init__

erohmensing:
So if it went 
Priority 1: Sklearn, Keras, Embedding
Priority 2: Memoization (also Augmented Memo)
Priority 3: Fallback, TwoStageFallback
Priority 4: Form

would that look right? That was the original plan reflecting the constants we're getting rid of

akelad:
and i assume a higher priority number means it's more important?

erohmensing:
correct, but only in the case of equal confidence, i.e. memoization will still beat out fallback  (regardless of its higher priority) if `not should_nlu_fallback` and there's a memorized action. if `should_nlu_fallback` then fallback will win, since it sets its highest confidence to 1.0 instead of `core_threshold`

rasa_core/policies/ensemble.py
```diff
@@ -55,6 +57,17 @@ def _training_events_from_trackers(training_trackers):
 
         return events_metadata
 
+    def _check_priorities(self) -> None:
+        priorities = []
```
```suggestion
        priorities = set()
```

rasa_core/policies/ensemble.py
```diff
@@ -55,6 +57,17 @@ def _training_events_from_trackers(training_trackers):
...
+        for p in self.policies:
+            if p.priority not in priorities:
+                priorities.append(p.priority)
+            else:
+                logger.warn("Multiple policies with the same priority " +
+                            "found in PolicyEnsemble. When personalizing "
+                            "priorities, be sure to give all policies "
+                            "different priorities.")
```
tmbo: please try to help the user as much as possible: e.g. link to docs, which priority is duplicated, which policy

rasa_core/policies/ensemble.py
```diff
@@ -315,17 +328,30 @@ def probabilities_using_best_policy(self,
...
             confidence = np.max(probabilities)
+
             if confidence > max_confidence:
                 max_confidence = confidence
                 result = probabilities
                 best_policy_name = 'policy_{}_{}'.format(i, type(p).__name__)
+                best_policy_priority = p.priority
+
+            # higher priority wins when confidence equal
+            elif confidence == max_confidence:
```
tmbo: this code should not be duplicated, should rather be an or in the `if` in L341

rasa_core/policies/form_policy.py
```diff
@@ -19,12 +18,14 @@ class FormPolicy(MemoizationPolicy):
 
     def __init__(self,
                  featurizer: Optional[TrackerFeaturizer] = None,
+                 priority: int = 4,
```
tmbo: why does the fallback persist the priority but the form does not?

rasa_core/policies/policy.py
```diff
@@ -38,14 +48,20 @@ def _load_tf_config(config: Dict[Text, Any]) -> Optional[tf.ConfigProto]:
...
         self.__featurizer = self._create_featurizer(featurizer)
+        self.__priority = self._set_priority(priority)
 
     @property
     def featurizer(self):
         return self.__featurizer
 
+    @property
+    def priority(self):
+        return self.__priority
```
tmbo: I don't actually see a reason why this should not just be a normal property

rasa_core/policies/policy.py
```diff
@@ -38,14 +48,20 @@ def _load_tf_config(config: Dict[Text, Any]) -> Optional[tf.ConfigProto]:
             return None
 
     def __init__(self,
-                 featurizer: Optional[TrackerFeaturizer] = None
+                 featurizer: Optional[TrackerFeaturizer] = None,
+                 priority: int = None
```
tmbo:
```suggestion
                 priority: Optional[int] = None
```

docs/changelog.rst
```diff
@@ -1,6 +1,10 @@
 :desc: Rasa Core Changelog
 
 .. include:: ../CHANGELOG.rst
+Added
+-------
+- Added `priority` property of policies to influence best policy in the case of equal confidence
```
tmbo: need to add to migration that one needs to pass priority property to the super class call if user wrote their own policy

rasa_core/policies/ensemble.py
```diff
@@ -55,6 +57,17 @@ def _training_events_from_trackers(training_trackers):
 
         return events_metadata
 
+    def _check_priorities(self) -> None:
```
tmbo: function should get a comment

rasa_core/policies/policy.py
```diff
@@ -29,6 +29,16 @@ def _create_featurizer(cls, featurizer=None):
         else:
             return cls._standard_featurizer()
 
+    @classmethod
+    def _set_priority(cls, priority=None):
+        if priority:
+            return copy.deepcopy(priority)
+        else:
+            logger.warning("Policy should have a default priority defined " +
```
tmbo: I think the default should be `1` and not `0` - any policy that does not set a priority should be treated like keras and embedding policies (this will also make this somewhat more backwards compatible, as this is how we treated them before)

tmbo: I don't even think we should have this warning. The policy base class should set it to a sensible default, if the user writes his custom policy I don't think we need to force them to define a priority.

rasa_core/policies/sklearn_policy.py
```diff
@@ -188,7 +195,13 @@ def load(cls, path: Text) -> Policy:
             ("Loaded featurizer of type {}, should be "
              "MaxHistoryTrackerFeaturizer.".format(type(featurizer).__name__))
 
-        policy = cls(featurizer=featurizer)
+        meta_file = os.path.join(path, "sklearn_policy.json")
+        if os.path.isfile(meta_file):
```
tmbo: why should this file not exist?

rasa_core/policies/two_stage_fallback.py
```diff
@@ -41,9 +41,11 @@ class TwoStageFallbackPolicy(FallbackPolicy):
     """
 
     def __init__(self,
+                 priority: int = 3,
                  nlu_threshold: float = 0.3,
                  core_threshold: float = 0.3,
-                 fallback_core_action_name: Text = ACTION_DEFAULT_FALLBACK_NAME,
+                 fallback_core_action_name:
```
tmbo: do we need this line break?

tests/test_ensemble.py
```diff
@@ -31,6 +35,76 @@ def test_policy_loading_simple(tmpdir):
...
+    def train(self, training_trackers, domain, **kwargs):
+        pass
+
+    def predict_action_probabilities(self, tracker, domain):
+        result = [0.0] * domain.num_actions
+        result[0] = 1.0
+        return result
+
+
+class Priority_2_Policy(Policy):
```
tmbo: I think this should not be duplicated, rather the above policy shoud be called `ConstantPolicy` or something like that and get the index to predict as an init param.

rasa_core/policies/two_stage_fallback.py
erohmensing: my editor was yelling at me because it was 80 instead of 79, not sure what our default is, because it probably will pass at 80 (because it obviously passed before)

rasa_core/policies/ensemble.py
erohmensing: so then should it look like this?
```
if (confidence > max_confidence or (confidence == max_confidence and p.priority > best_policy_priority)
```

tmbo: i think this should work as well
```python
if (confidence, p.priority) > (max_confidence, best_policy_priority):
```

erohmensing: Yep, tested it and it does have the functionality we want, i can implement that. How does that syntax work exactly? Does it compare the first and if they're equal move on to comparing the second? I haven't seen this and am having trouble looking it up

tmbo: it is pythons normal way to compare tuples

rasa_core/policies/form_policy.py
erohmensing: form_policy doesn't have its own persist method since its a subclass of memo, it persists via memo's persist method, which persists priority. With regards to memo's persisting, I wasn't sure if I should create a new meta file to parallel the others, but it seemed like metadata was already being persisted via the `data`  file

rasa_core/policies/sklearn_policy.py
erohmensing: Ah, I didn't notice that sklearn_policy's `model` doesn't default to `None` like keras_policy's does, so it'll always persist something. Unless you specifically instantiate your sklearn_policy with `model=None`, overriding the default model (I admit there is a super slim chance of someone doing this)

Isn't this super slim chance the only possible way to hit the warning in `sklearn_policy.persist()` though?

```
    def persist(self, path: Text) -> None:

        if self.model:
            self.featurizer.persist(path)

            filename = os.path.join(path, 'sklearn_model.pkl')
            with open(filename, 'wb') as f:
                pickle.dump(self._state, f)
        else:
            warnings.warn("Persist called without a trained model present. "
                          "Nothing to persist then!")
```

docs/changelog.rst
erohmensing: only if they want to define a priority right? by mixing your suggestions I was thinking `policy.__init__()` can include `priority: Optional[int] = 1`

tmbo: yes, if we have a sane default prio, no migration is necessary

rasa_core/policies/ensemble.py
```diff
@@ -55,6 +57,25 @@ def _training_events_from_trackers(training_trackers):
...
+        """Checks for duplicate policy priorities within PolicyEnsemble."""
+
+        priority_dict = {}
```
tmbo: add import `from collections import defaultdict`

```suggestion
        priority_dict = defaultdict(list)
```

rasa_core/policies/ensemble.py
```diff
@@ -55,6 +57,25 @@ def _training_events_from_trackers(training_trackers):
...
+        for p in self.policies:
+            if p.priority in priority_dict:
```
tmbo: remove all the if / else (L65-L68) and replace with
`priority_dict[p.priority].append(type(p).__name__)`
(the defaultdict will automatically create a list if it is not already there, so we do not need to check that)

rasa_core/policies/ensemble.py
```diff
@@ -55,6 +57,25 @@ def _training_events_from_trackers(training_trackers):
...
+                priority_dict[p.priority].append(type(p).__name__)
+            else:
+                priority_dict[p.priority] = [type(p).__name__]
+
+        for k, v in priority_dict.items():
+            if len(v) > 1:
+                logger.warn(("Found policies {} with same priority {} "
```
tmbo: (`warn` is deprecated, make sure indentation of the following lines is correct)
```suggestion
                logger.warning(("Found policies {} with same priority {} "
```

rasa_core/policies/ensemble.py
erohmensing: ah dope, I should have assumed when i ran into that problem that python already had it figured out :)

tmbo: no worries, that's what a code review is for - to spread the word :wink:

PR #1740, opened 2019-02-19
fork saving fix

### LINKED ISSUE #1606: Forks not saved in interactive learning
**Rasa Core version**:
0.12.4
**Python version**: 
3.6
Ubuntu
**Issue**:
When I run an interactive learning session to train longer stories with different decisions, it would be useful to be able to use the Fork feature to generate different training stories. However, atm only the current / latest fork at the time of the export is stored in the stories.md. All the others are lost.

paulaWesselmann:
**Proposed changes**:
- saves story when fork is created
- fixes #1606 

- [ ] updated the changelog

### DIFF
--- a/rasa_core/training/interactive.py
+++ b/rasa_core/training/interactive.py
@@ -588,6 +588,17 @@ def _ask_if_quit(sender_id: Text, endpoint: EndpointConfig) -> bool:
     elif answer == "undo":
         raise UndoLastStep()
     elif answer == "fork":
+        # a fork is created and the first story line saved as if quit
+        story_path, nlu_path, domain_path = _request_export_info()
+
+        tracker = retrieve_tracker(endpoint, sender_id)
+        evts = tracker.get("events", [])
+
+        _write_stories_to_file(story_path, evts)
+        _write_nlu_to_file(nlu_path, evts)
+        _write_domain_to_file(domain_path, evts, endpoint)
+
+        logger.info("Successfully wrote stories and NLU data")
         raise ForkTracker()
     elif answer == "restart":
         raise RestartConversation()
--- a/rasa_core/training/interactive.py
+++ b/rasa_core/training/interactive.py
@@ -555,7 +555,7 @@ def _slot_history(tracker_dump: Dict[Text, Any]) -> List[Text]:
     return slot_strs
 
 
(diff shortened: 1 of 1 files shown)


rasa_core/training/interactive.py
```diff
@@ -588,6 +588,17 @@ def _ask_if_quit(sender_id: Text, endpoint: EndpointConfig) -> bool:
...
+        story_path, nlu_path, domain_path = _request_export_info()
+
+        tracker = retrieve_tracker(endpoint, sender_id)
+        evts = tracker.get("events", [])
+
+        _write_stories_to_file(story_path, evts)
+        _write_nlu_to_file(nlu_path, evts)
+        _write_domain_to_file(domain_path, evts, endpoint)
+
+        logger.info("Successfully wrote stories and NLU data")
```
akelad: since this whole chunk of code is used in two places now, I'd suggest we make a function and call that instead

rasa_core/training/interactive.py
```diff
@@ -555,6 +555,21 @@ def _slot_history(tracker_dump: Dict[Text, Any]) -> List[Text]:
     return slot_strs
 
 
+def _write_story_write_nlu(sender_id: Text, endpoint: EndpointConfig):
```
akelad: let's call the function `_write_data_to_file` so it's clearer

rasa_core/training/interactive.py
```diff
@@ -588,6 +594,8 @@ def _ask_if_quit(sender_id: Text, endpoint: EndpointConfig) -> bool:
     elif answer == "undo":
         raise UndoLastStep()
     elif answer == "fork":
+        # a fork is created and the first story line saved as if quit
+        _write_story_write_nlu(sender_id, endpoint)
```
tmbo: I do not think this is a good idea from the usability perspective: this will require the user to enter the export info every time they fork a conversation. rather I'd suggest to use the `sender_ids` that we are collecting anyways in `record_messages` to dump all the conversation at once on quit.

akelad: good point, didn't think of that one

paulaWesselmann: I'll change that

PR #1743, opened 2019-02-20
Rasa cli

wochinge:
**Proposed changes**:
- changes for https://github.com/rasahq/rasa_stack/issues/4 and https://github.com/rasahq/rasa_stack/issues/1
- utility functions for colored logging
- open webbrowser when visualizing stories
- renamed ``train_dialogue_model`` to ``train``
- renamed ``rasa_core.evaluate`` to ``rasa_core.test``
- changed export paths for interactive training to ``data/core/stories.md``,
  ``data/nlu/nlu.md`` and ``data/nlu/nlu_interactive.md``

- ~[ ] added some tests for the functionality~
- [x] updated the changelog


wochinge:
@tmbo Ready for review. I rely on the master version of NLU though, so the travis built won't pass.

rasa_core/evaluate.py
```diff
@@ -1,667 +1,10 @@
-from collections import defaultdict
```
tmbo: This def needs to go into the migration guide

rasa_core/config.py
```diff
@@ -3,14 +3,14 @@
 from typing import Optional, Text, List
 
 from rasa_core import utils
-from rasa_core.policies import PolicyEnsemble
```
tmbo: I guess you moved these to make the commandline faster. 

Does it make sense to add a test to travis that checks how long it takes for the cli to start?

rasa_core/training/interactive.py
```diff
@@ -1062,7 +1062,7 @@ def _enter_user_message(sender_id: Text,
                         endpoint: EndpointConfig) -> None:
     """Request a new message from the user."""
 
-    question = questionary.text("Next user input (Ctr-c to abort):")
+    question = questionary.text("Your input ->")
```
tmbo: I think we should keep a message about how to exit (we can print that separately though)

rasa_core/training/interactive.py
```diff
@@ -1317,7 +1317,7 @@ def visualisation_png():
             abort(404)
 
 
-def run_interactive_learning(agent: Agent,
+def run_interactive_learning(agent: 'Agent',
```
tmbo: This shouldn't be a string. The only valid location for a type annotation to be a string is when the method is defined in the same module before the class:
```
def f(a: 'A') -> None:
   pass

class A:
  pass
```
Everywhere else, the type annotation should be the classname. If you want to prevent it from being imported on runtime, put it into an `if typing.TYPE_CHECKING:` at the top
```suggestion
def run_interactive_learning(agent: Agent,
```

rasa_core/training/visualization.py
```diff
@@ -32,8 +31,8 @@ def __init__(self, nlu_training_data):
 
     @staticmethod
     def _create_reverse_mapping(
-        data: TrainingData
-    ) -> Dict[Dict[Text, Any], List[Message]]:
+        data: 'TrainingData'
```
tmbo: same here, not going to comment on all of them 😅

rasa_core/utils.py
```diff
@@ -549,11 +536,28 @@ def read_lines(filename, max_line_limit=None, line_pattern=".*"):
                 break
 
 
+def file_as_bytes(path: Text) -> bytes:
+    """Read in a file as a byte array."""
+    with io.open(path, 'rb') as f:
```
tmbo: should we add encoding utf-8?

requirements.txt
```diff
@@ -17,7 +17,7 @@ coloredlogs==10.0
 flask==1.0.2
 flask_cors==3.0.7
 scikit-learn==0.20.2
-rasa_nlu~=0.14.0
+rasa_nlu~=0.15.0a1
```
tmbo: well, merge nlu first and then tag an a1 on nlu

setup.py
```diff
@@ -58,8 +58,8 @@
     "gevent~=1.4",
     "pytz~=2018.9",
     "python-dateutil~=2.7",
-    "rasa_nlu~=0.14.0",
-    "rasa_core_sdk~=0.12.1",
+    "rasa_nlu~=0.15.0a1",
+    "rasa_core_sdk~=0.13.0a1",
```
tmbo: same here

CHANGELOG.rst
```diff
@@ -16,11 +16,17 @@ Added
...
+- utility functions for colored logging
+- open webbrowser when visualizing stories
 
 Changed
 -------
 - starter packs are now tested in parallel with the unittests,
   and only on master and branches ending in ``.x`` (i.e. new version releases)
+- renamed ``train_dialogue_model`` to ``train``
+- renamed ``rasa_core.evaluate`` to ``rasa_core.test``
+- changed export paths for interactive training to ``data/core/stories.md``,
```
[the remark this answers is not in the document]
wochinge: how would we then distinguish the files?

rasa_core/agent.py
```diff
@@ -348,7 +337,7 @@ def handle_text(
         message_preprocessor: Optional[Callable[[Text], Text]] = None,
         output_channel: Optional[OutputChannel] = None,
         sender_id: Optional[Text] = UserMessage.DEFAULT_SENDER_ID
-    ) -> Optional[List[Text]]:
+    ) -> Optional[List[Any]]:
```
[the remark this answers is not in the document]
wochinge: it can also returns lists of dicts, not sure what other types are possible

rasa_core/training/interactive.py
wochinge: where would you print that? we can also keep the `ctr-c` thing and also add it to the scaffold at the cli

rasa_core/training/interactive.py
wochinge: i think the problem was that the command line interface somehow had the type checking activated

rasa_core/training/interactive.py
tmbo: I'd print that right at the beginning but maybe it is not to important and users just now how this works, @JustinaPetr thoughts on this?

rasa_core/agent.py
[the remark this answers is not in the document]
tmbo: mhm you are right it returns dict, but I think if something is returned it is always the collected messages from the output channel, which is a list of dicts, so I guess `Optional[List[Dict[Text, Any]]]`

rasa_core/config.py
wochinge: added reminder to stack pr

rasa_core/training/interactive.py
wochinge: @JustinaPetr

CHANGELOG.rst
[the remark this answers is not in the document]
wochinge: ok, then we put this in `data`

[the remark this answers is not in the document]
wochinge: :+1:

rasa_core/training/interactive.py
JustinaPetr: Sorry for the late response. I would say we should keep it printed, because it's not really clear how to exit. Ctr-c is an obvious choice by habit, but to have a good UX, I would suggest having it printed. RIght after beginning looks like a great place for that.

wochinge: Okay cool! that's already done :-)

wochinge:
@tmbo Quick note: I had to change the test discovery to `test_` since pytest otherwise tried to execute the `test` method of `rasa_core.test`: https://github.com/RasaHQ/rasa_core/pull/1743/commits/1fc2bdd5b2f574086a1391b6ba275fbb279810e8

tmbo:
Yes sounds good :+1:

PR #1761, opened 2019-02-25
Interactive

### LINKED ISSUE #1749: Entity prompt doesn't appear in interactive training when there are no entities recognised
in interactive learning sometimes an intent is correctly identified, but includes an entity that is not annotated, only when the intent is marked as wrong, but then again is chosen as correct intent the user is asked to annotate the entity

paulaWesselmann:
**Proposed changes**:
- changed to ask the user to verify entities as well as intent in interactive learning, fixes #1749 

- [ ] updated the changelog


akelad:
@paulaWesselmann is this different to your fork saving fix? looks like your commits from that are included here too

paulaWesselmann:
@akelad yes, sorry I realized that too. The fork changes are already approved by Tom. the only thing else that changed here is the print statement for the user in interactive learning to ask for intent and entity.

rasa_core/training/interactive.py
```diff
@@ -1005,6 +1005,7 @@ def _validate_user_text(latest_message: Dict[Text, Any],
     else:
         question = questionary.confirm(
             "Is the NLU classification for '{}' with intent '{}' correct?"
+            "\n    AND are all occurring entities labeled correctly?"
```
akelad: could we maybe rephrase this so it's a little more intuitive? Right now the sentences together don't make much sense. So, phrasing it like "Is the NLU classification for '{}': intent - {} and entities: {} correct?". Where entities would be none in the case of none of them

rasa_core/training/interactive.py
```diff
@@ -997,15 +997,23 @@ def _validate_user_text(latest_message: Dict[Text, Any],
...
         print("The NLU classification for '{}' returned '{}'"
               "".format(text, intent))
         return False
     else:
         question = questionary.confirm(
-            "Is the NLU classification for '{}' with intent '{}' correct?"
-            "".format(text, intent))
+            "Is the NLU classification for '{}' with intent '{}' "
+            "and entities: {} correct?"
+            "".format(text, intent, entity))
```
akelad: ahh sorry about the back and forth on this, but this looks weird now:
` Is the NLU classification for 'i want a [greek](cuisine) restaurant' with intent 'inform' and entities: ['[greek](cuisine)'] correct?`

Can you come up with a better phrasing for this? maybe post it here and we can discuss before you change it again. Something like asking are all marked entities correct and the intent classification

paulaWesselmann: Do you think it is better to make this a two step question? first if the intent is classified correctly and then the entity? Or is that too many questions for the user?

akelad: hmmm i'm not sure which is better now tbh, could you post this in slack and ask for input? I feel like the two step thing might be too many times to hit enter

paulaWesselmann:
I changed it to what we discussed in slack. It's ready for (hopefully) final review now.

rasa_core/training/interactive.py
```diff
@@ -999,21 +999,20 @@ def _validate_user_text(latest_message: Dict[Text, Any],
     intent = parse_data.get("intent", {}).get("name")
     entities = parse_data.get("entities", [])
     if entities:
-        entity = []
-        for e in entities:
-            entity.append('[' + e['value'] + '](' + e['entity'] + ')')
+        msg = ("Is the intent '{}' correct for '{}' and are "
```
akelad: can we just rename `msg` to `message` everywhere?

PR #1762, opened 2019-02-26
update interactive.py

yerok:
When trying to detect format of the training data file in the `_write_nlu_to_file` function, there is a conflict between the opening of the file in writing mode and the call to the `_guess_format` fonction from rasa_nlu/rasa_nlu/training_data/loading.py.

Indeed, after the opening, the file is everytime detected as markdown by the `_guess_format` func because the `js = json.loads(content)` loads an empty file resulting in wrong detection (if the file is a json). 

Quick fix to avoid that.

**Proposed changes**:
- opening the file in writing mode only after detecting the format

- [ ] updated the changelog

### DIFF
--- a/rasa_core/training/interactive.py
+++ b/rasa_core/training/interactive.py
@@ -739,13 +739,17 @@ def _write_nlu_to_file(
         previous_examples = TrainingData()
 
     nlu_data = previous_examples.merge(TrainingData(msgs))
-
+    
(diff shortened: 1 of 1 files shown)


akelad:

tmbo:
Thanks a lot for your contribution - I merged the commits but through a separate PR (there are some style issues with yours that result in the travis failure).

PR #1807, opened 2019-03-08
Stack training endpoint

### LINKED ISSUE #1806: Rasa stack training endpoint

ricwo:
**Proposed changes**:
- fixes #1806 

- [x] updated the changelog


CHANGELOG.rst
```diff
@@ -25,6 +25,7 @@ Added
 - added ability to use multiple env vars per line in yaml files
 - added ``priority`` property of policies to influence best policy in 
   the case of equal confidence
+- Rasa Stack HTTP training endpoint at ``POST /jobs``
```
tmbo:
```suggestion
- Rasa Stack HTTP training endpoint at ``POST /jobs``. This endpoint 
  will train a combined Rasa Core and NLU model.
```

data/test_nlu_data/nlu.md
```diff
@@ -0,0 +1,50 @@
+## intent:greet
```
tmbo: do we need to duplicate this, seems to be an exact copy of the moodbot data

docs/_static/spec/server.yml
```diff
@@ -624,6 +624,39 @@ paths:
...
           $ref: '#/components/responses/400Evaluation'
 
+  /jobs:
+    post:
+      security:
+      - TokenAuth: []
+      - JWT: []
+      tags:
+      - Model
+      summary: Train Rasa Stack model
```
tmbo:
```suggestion
      summary: Train a Rasa Stack model
```

docs/_static/spec/server.yml
```diff
@@ -624,6 +624,39 @@ paths:
...
+      description: >-
+        Trains a Rasa Stack model
```
tmbo: don't be so lazy 😛
```suggestion
        Trains a Rasa Stack model. A stack model is a model combining a trained dialogue model with an NLU model.
```

rasa_core/agent.py
```diff
@@ -704,5 +713,5 @@ def _form_policy_not_present(self) -> bool:
             if there is a form action in the domain
         """
         return (self.domain and self.domain.form_names and not
-                any(isinstance(p, FormPolicy)
-                    for p in self.policy_ensemble.policies))
+        any(isinstance(p, FormPolicy)
```
tmbo: I don't think pep8 will like this 😅

rasa_core/run.py
```diff
@@ -1,5 +1,6 @@
 import argparse
 import logging
+import sys
```
tmbo:
```suggestion
```

rasa_core/server.py
```diff
@@ -508,8 +507,24 @@ def evaluate_stories():
...
-    @app.route("/domain",
-               methods=['GET', 'OPTIONS'])
+    @app.route("/jobs", methods=['POST', 'OPTIONS'])
+    @requires_auth(app, auth_token)
+    @cross_origin(origins=cors_origins)
+    def train_stack():
+        from rasa.cli.train import train
+        from argparse import Namespace
+
+        """Train a Rasa Stack model."""
```
tmbo: should be below the function `def`

rasa_core/server.py
```diff
@@ -508,8 +507,24 @@ def evaluate_stories():
...
+        args = Namespace(**request.get_json())
+        print("got args", args, request.get_json())
```
tmbo:
```suggestion
```

rasa_core/server.py
```diff
@@ -508,8 +507,24 @@ def evaluate_stories():
...
+        try:
+            model_name = train(args)
+            return jsonify({'info': 'new model trained', 'model': model_name})
```
tmbo: I don't think this is how this should work: we need this function when running on the server as well to train models in a worker. In that case the training data and the domain file won't be on the worker server, but rather should be transmitted with the request, as we do it in Rasa NLU (where we post the config and the training data to the endpoint - not only file names).

In addition to that, I think this should return the model instead of json

requirements.txt
```diff
@@ -41,3 +41,4 @@ python-socketio==3.1.1
 pydot==1.4.1
 keras-applications==1.0.6
 keras-preprocessing==1.0.5
+rasa~=0.2.0a3
```
tmbo: this will create a cyclic dependency - are you sure the installation still works? 😅  might be better to just assume it is installed and error if the package can not be found.

tmbo: but not too important as it will be fixed by core & nlu merge but still want to make sure install still works.

tmbo: If helpful, we can also already merge the rasa_stack code in the core repository

rasa_core/agent.py
ricwo: true! pycharm auto-indent 😄

data/test_nlu_data/nlu.md
ricwo: you're right, we can just use that file

rasa_core/server.py
ricwo: hmm, thought it was a convention because it's the same in `evaluate_stories()` above. I'll change both

requirements.txt
ricwo: yes, I thought of this as an intermediary solution until the repos are merged (and yes installation works - I needed this for the tests)

tmbo: :+1:

tmbo: now that the repositories are merged, let's remove this requirement :tada:

ricwo:
@tmbo this is ready for another review

dev-requirements.txt
```diff
@@ -10,6 +10,7 @@ treq==17.8.0
 freezegun==0.3.10
 responses==0.9.0
 httpretty==0.9.5
+sklearn-crfsuite==0.3.6
```
tmbo: I think this is already part of the `requirements.txt`

tests/test_server.py
```diff
@@ -249,6 +252,33 @@ def test_evaluate(app):
...
 
 
+def test_stack_training(app,
+                        default_domain_path,
+                        default_stories_file,
+                        default_stack_config,
+                        default_nlu_data
+                        ):
+    payload = dict(
+        domain=open(default_domain_path).read(),
```
tmbo: let's make sure these file handles are closed - it is a bit more verbose, but avoids warnings about unclosed handles:
```
domain_file = open(default_domain_path)
payload = ...
domain_file.close()
```

Issue #1732, opened 2019-02-18
Interactive Learning: Only include ner_crf entities in nlu export by including the extractor field in each entity at NLU


wochinge:
At the end of interactive learning you can save stories and NLU data which you collected during the training. However, the exported NLU data should only include entity annotation for entities which were 
* extracted by ner_crf
* manually annotated by the user

akelad:
@wochinge we could export the ones recognised by duckling as well, as long as the `extractor` field is correct right?

paulaWesselmann:
@wochinge why should it only include entities extracted by ner_crf and manually annotated? And which other entities does it extract currently?  In my experiments I can only see that it additionally picks up capital spellings of words as synonym of the lower case word and records this in NLU as well.

paulaWesselmann:
#1725

paulaWesselmann:
Ok, got it from #1725

wochinge:
@akelad Why should we also export the things from Duckling? We don't want train ner_crf on it right?

akelad:
well, we won't, if there's an `extractor: ner_duckling` field present, it will skip it. but actually that doesn't work for markdown, only for json :D so yeah better not export it at all

wochinge:
Good hint! I am a post-json Rasa user so I have no idea what's included in the json format :-D

akelad:
the rasa nlu json format actually does still exist :D

PR #1814, opened 2019-03-11
Inter entity

### LINKED ISSUE #1728: Correcting the intent in interactive learning mode removes additional entity information
**Rasa Core version**: 0.13.2
**Python version**: 3.6.8
**Operating system**: windows

When one corrects the intent in the interactive learning mode, but leaves the entities untouched they get altered.

Look at this example:
`Is the NLU classification for '[Am Mittwoch](time:2019-02-20T00:00:00.000+01:00)' with intent 'choose_appointment'`

The date was recognized by duckling. Without altering the intent the entity would get saved as following in the tracker:

```
"entities": [
            {
              "start": 0,
              "end": 11,
              "text": "Am Mittwoch",
              "value": "2019-02-20T00:00:00.000+01:00",
              "confidence": 1,
              "additional_info": {
                "values": [
                  {
                    "value": "2019-02-20T00:00:00.000+01:00",
                    "grain": "day",
                    "type": "value"
                  },
                  {
                    "value": "2019-02-27T00:00:00.000+01:00",
                    "grain": "day",
                    "type": "value"
                  },
                  {
                    "value": "2019-03-06T00:00:00.000+01:00",
                    "grain": "day",
                    "type": "value"
                  }
                ],
                "value": "2019-02-20T00:00:00.000+01:00",
                "grain": "day",
                "type": "value"
              },
              "entity": "time",
              "extractor": "ner_duckling_http"
            }
          ],
```

After altering the recognized intent, but leaving the entity untouched, it gets reduced to:

```
"entities": [
              {
                "end": 11,
                "entity": "time",
                "start": 0,
                "value": "2019-02-20T00:00:00.000+01:00"
              }
            ],
```

paulaWesselmann:
**Proposed changes**:
- fixes #1728 when updating entities in interactive learning only annotated entities are updated

- [x] updated the changelog


paulaWesselmann:
Note that this PR includes the not yet merged changes for #1732.

CHANGELOG.rst
```diff
@@ -30,6 +30,7 @@ Changed
 -------
 - starter packs are now tested in parallel with the unittests,
   and only on master and branches ending in ``.x`` (i.e. new version releases)
+- for IL only include manually annotated and ner_crf entities in nlu export
```
tmbo:
```suggestion
- for interactive learning only include manually annotated and ner_crf entities in nlu export
```

rasa_core/training/interactive.py
```diff
@@ -686,6 +686,15 @@ def _collect_messages(evts: List[Dict[Text, Any]]) -> List[Message]:
     for evt in evts:
         if evt.get("event") == UserUttered.type_name:
             data = evt.get("parse_data")
+            logger.info("Those are the collected entities:" + str(
```
tmbo: I think we do not need this logger stmt

rasa_core/training/interactive.py
```diff
@@ -686,6 +686,15 @@ def _collect_messages(evts: List[Dict[Text, Any]]) -> List[Message]:
...
+                data["entities"]))
+            for entity in data["entities"]:
+                if "extractor" in entity:
```
tmbo: in which cases is this missing from the entitiy?

rasa_core/training/interactive.py
```diff
@@ -686,6 +686,15 @@ def _collect_messages(evts: List[Dict[Text, Any]]) -> List[Message]:
...
+                    # only inlcude entities which are extracted by ner_crf
+                    # or manually annotated (therefore lack entity["extractor"])
+                    if entity["extractor"] != 'CRFEntityExtractor':
```
tmbo: this should rather use `CRFEntityExtractor.__name__` to avoid a soft reference that can not be tracked by e.g. IDEs

tmbo: Why do we only include entities from the CRF, what if the user wrote his own entity component?

rasa_core/training/interactive.py
```diff
@@ -1060,21 +1069,46 @@ def _validate_nlu(intents: List[Text],
 
 def _correct_entities(latest_message: Dict[Text, Any],
                       endpoint: EndpointConfig,
-                      sender_id: Text) -> Dict[Text, Any]:
+                      sender_id: Text) -> [Dict[Text, Any]]:
```
tmbo: this is not a proper type
```suggestion
                      sender_id: Text) -> List[Dict[Text, Any]]:
```

rasa_core/training/interactive.py
```diff
@@ -1060,21 +1069,46 @@ def _validate_nlu(intents: List[Text],
...
-    entity_str = _as_md_message(latest_message.get("parse_data", {}))
+    parse_data = latest_message.get("parse_data", {})
+    entity_str = _as_md_message(parse_data)
     question = questionary.text(
         "Please mark the entities using [value](type) notation",
         default=entity_str)
 
     annotation = _ask_or_abort(question, sender_id, endpoint)
     # noinspection PyProtectedMember
     parsed = MarkdownReader()._parse_training_example(annotation)
```
tmbo: please rename the variables, now there is `parse_data` and `parsed`

rasa_core/training/interactive.py
```diff
@@ -1060,21 +1069,46 @@ def _validate_nlu(intents: List[Text],
...
-    return parsed.get("entities", [])
+    if entity_str == parsed:  # user did not change annotation
+        return parse_data.get("entities", [])
+
+    else:  # update annotated entities
+        entities = []
```
tmbo: This code is way to hard to read. I tried to simplify it, and this should roughly do the same thing, right?

```python
        entities = parsed.get("entities", [])[:]

        # overwrite entities which have already been 
        # annotated in the original annotation to preserve
        # additional entity parser information
        for i, entity in enumerate(entities):
            for original_entity in parse_data.get("entities", []):
                if _is_same_entity_annotation(entity, original_entity):
                    entities[i] = original_entity
                    break
        return entities

# adding this function as well
def _is_same_entity_annotation(entity, other):
    return (entity['value'] == other['value'] and
            entity['entity'] == other['entity'])

```
I also don't think we need the if else in L1087 (it should be fine to always run the above, even if the user did not change anything)

rasa_core/training/interactive.py
paulaWesselmann: Whenever the user manually annotated the entity

rasa_core/training/interactive.py
paulaWesselmann: See issue #1725. If the user wrote his own entity component the easiest solution would be if they just do not add an anything for `entity["extractor"]` depending on if they want it to be included in the training stories or not. Or we have to filter out the extractors we know never should be included here and if the user writes a new component it will by default be included in the stories.

rasa_core/training/interactive.py
tmbo: makes sense, let's not use two if's though and just go with `if entity.get("extractor") != ...`

rasa_core/training/interactive.py
tmbo: let's do it the other way around and test for the ones that we should not include as that is the way we do it during the NLU model evaluation where we only evaluate the processors that can be trained

rasa_core/training/interactive.py
```diff
@@ -686,6 +690,21 @@ def _collect_messages(evts: List[Dict[Text, Any]]) -> List[Message]:
...
             data = evt.get("parse_data")
+
+            for entity in data["entities"]:
+
+                excluded_extractors = [
+                    DucklingHTTPExtractor.__name__,
+                    SpacyEntityExtractor.__name__,
+                    MitieEntityExtractor.__name__
+                ]
+                logger.info("Exclude entity marking of following extractors {} "
```
tmbo:
```suggestion
                logger.debug("Exclude entity marking of following extractors {} "
```

rasa_core/training/interactive.py
```diff
@@ -686,6 +690,21 @@ def _collect_messages(evts: List[Dict[Text, Any]]) -> List[Message]:
     for evt in evts:
         if evt.get("event") == UserUttered.type_name:
             data = evt.get("parse_data")
+
+            for entity in data["entities"]:
```
tmbo:
```suggestion
            for entity in data.get("entities", []):
```

PR #1827, opened 2019-03-13
Added Rasa Stack files

tmbo:
**Proposed changes**:
- moved files over from https://github.com/RasaHQ/rasa_stack

- [x] updated the changelog


rasa_core/server.py
```diff
@@ -183,7 +183,7 @@ def create_app(agent,
     @cross_origin(origins=cors_origins)
     def hello():
         """Check if the server is running and responds with the version."""
-        return "hello from Rasa Core: " + __version__
+        return "hello from Rasa Core: " + rasa.__version__
```
wochinge: "hello from rasa core" does not make really sense then right?

setup.py
```diff
@@ -67,15 +67,16 @@
     "python-socketio~=3.0",
     "pydot~=1.4",
     "keras-applications==1.0.6",
-    "keras-preprocessing==1.0.5"
+    "keras-preprocessing==1.0.5",
+    "sklearn-crfsuite~=0.3.6",
```
wochinge: rasa_nlu >= 0.15.0a2

wochinge: same for requirements.txt

rasa_core/server.py
```diff
@@ -9,14 +9,14 @@
 from flask_cors import CORS, cross_origin
 from flask_jwt_simple import JWTManager, view_decorators
 
+import rasa
```
wochinge: is this also in python sys path if you build & run the docker image?

tmbo: yes, if you run `pip install rasa` (or `pip install -e .`) it will install two packages, `rasa` and `rasa_core`

setup.py
tmbo: requirements is already on `rasa_nlu~=0.15.0a4`

PR #1838, opened 2019-03-14
restructure training and policies and fallback actions docs

### LINKED ISSUE #1678: Update Training and Policies in docs
Core docs still encourage you to pass policy parameters to agent.train(). These should be updated but I'm still working on implementing policy priority, so that should go in this update too.

erohmensing:
**Proposed changes**:
- fixes #1678 master changes (see issue for detailed list of changes)

Probably best to code review by pulling the branch and making the livedocs.

- [ ] updated the changelog


erohmensing:
@JustinaPetr sure, I can do that. It's happening because I decided to bold it instead of making it an actual header (as it is in the Embedding Policy) so that it wouldn't show up in the table of contents as I thought the TOC was perhaps getting a little too fleshed out. I see what you mean though by it looking like a new paragraph because of it being bigger, i'll change it to a header and see what you think.

erohmensing:
Actually, I just realized the Embedding Policy has everything tabbed in after the configuration, which makes a lot more sense. I'll try that first and see if that's enough to keep the flow

PR #1839, opened 2019-03-14
handle intent none

### LINKED ISSUE #1748: Exception in Fallback Policy if intent is None
**Rasa Core version**:
0.13.20

**Python version**: 
Python 3.6.8

Win 10

**Issue**:
https://github.com/RasaHQ/rasa_core/blob/master/rasa_core/policies/fallback.py#L97
```
  File "...Python36\lib\site-packages\rasa_core\policies\fallback.py", line 97, in predict_action_probabilities
AttributeError: 'NoneType' object has no attribute 'get'
```

if ```nlu_data["intent"]``` is None, an exception will be thrown. As this is a default policy, it should be robust enough to fail gracely or do something proper instead of failing at all.

We ran in this case with an untrained NLU model for a simple input like a postal code. Anyway, an exception is a quite harsh way to terminate and wasn't what we expected :-)

paulaWesselmann:
**Proposed changes**:
- fixes #1748 

- [ ] updated the changelog


rasa_core/policies/fallback.py
```diff
@@ -93,7 +93,11 @@ def predict_action_probabilities(self,
         # if NLU interpreter does not provide confidence score,
         # it is set to 1.0 here in order
         # to not override standard behaviour
-        nlu_confidence = nlu_data["intent"].get("confidence", 1.0)
+        if nlu_data["intent"]:
+            nlu_confidence = nlu_data["intent"].get("confidence", 1.0)
+        else:
+            logger.warning("NLU intend is None.")
```
tmbo: what do we need this logging message for?

paulaWesselmann: I wasn't sure if no intent is an issue that should be fixed. Since it seems to be normal for some policies should I remove this or make it a debug logging?

tmbo: yes, I think we can remove it :+1:

rasa_core/policies/fallback.py
```diff
@@ -93,7 +93,10 @@ def predict_action_probabilities(self,
         # if NLU interpreter does not provide confidence score,
         # it is set to 1.0 here in order
         # to not override standard behaviour
-        nlu_confidence = nlu_data["intent"].get("confidence", 1.0)
+        if nlu_data["intent"]:
+            nlu_confidence = nlu_data["intent"].get("confidence", 1.0)
+        else:
+            nlu_confidence = 1.0
```
erohmensing: Shouldn't this confidence be `0.0` since there's no recognized intent and we definitely want to be below the NLU threshold so we can NLU fallback? 

Also thinking this should maybe be try/except since it should happen only rarely

paulaWesselmann: I think in this case it is okay to predict fallback action with confidence `core_threshold` and not `1.0`. I don't think this changes the behavior here, but if the system learns to respond with a certain action for intent None, the confidence for this intent should be `1.0` and not `0.0`. Does that make sense?

PR #1841, opened 2019-03-14
Support unfeaturization of specific entities per intent

### LINKED ISSUE #1723: Support Unfeaturized Entities
Recognized entities can lead to prediction errors of Core although they might not be important for a current intent at all.

Currently, intents can only ignore all intents, but not only ignore a certain subset of intents.

Hence, it be nice if it would be possible to
- ignore certain entities for certain intents
- or mark certain entities as unfeaturized

### LINKED ISSUE #2984: Support Unfeaturized Entities
Recognized entities can lead to prediction errors of Core although they might not be important for a current intent at all.

Currently, intents can only ignore all intents, but not only ignore a certain subset of intents.

Hence, it be nice if it would be possible to
- ignore certain entities for certain intents
- or mark certain entities as unfeaturized

mauricedoepke:
fixes https://github.com/RasaHQ/rasa_nlu/issues/2984

- [ ] updated the changelog


mauricedoepke:
fixes #1700 as well

mauricedoepke:
It might be wise to annotate in interactive learning mode which entities are featurized and which not.

docs/domains.rst
```diff
@@ -220,7 +220,17 @@ file like this:
...
     - greet: {use_entities: false}
 
-This means that entities for those intents will be unfeaturized and therefore
+To ignore some entities or explicitly take only certain entities
+into account you can use this syntax:
+
+.. code-block:: yaml
+
+  intents:
+    - greet: {use_entities: false, include_entities: [name]}
```
wochinge: If one entity is explicitly included, we don't need the `use_entities: false` right?

data/test_domains/default_unfeaturized_entities.yml
```diff
@@ -0,0 +1,22 @@
+intents:
+ - greet: {use_entities: false, include_entities: [name]}
```
wochinge: `use_entities: false` is not explicitly needed, right?

CHANGELOG.rst
```diff
@@ -27,6 +27,7 @@ Added
 - added ``priority`` property of policies to influence best policy in 
   the case of equal confidence
 - added rasa command line interface and API
+- added support to indivually in/exclude entities for every intent
```
```suggestion
- added support to individually in-/exclude entities for every intent
```

wochinge (changes requested):
Left a few minor questions, but else it looks very good!
I really like your coding style :rocket:

docs/domains.rst
mauricedoepke: Currently `use_entities` defaults to true, no matter whether `include_entities` or `exclude_entities` are specified. Thus, if we want to include only one specific entity we need to exclude all by default, by setting `use_entities: false` and then explicitly include the wanted entity. Otherwise, all entities will get included.

But we could change the pr to implicitly set `use_entities: false` to false if `include_entities` is given.

wochinge: Mhm, how about changing it so that `include_entities: ["default"]` automatically means that all others are excluded? This seems more intuitive to me, than `use_entities: false` + `include_entities`.

mauricedoepke: I'll add that

mauricedoepke:
@tmbo @wochinge Do you handle the porting or do I have to do anything to prepare that?

mauricedoepke:
```
        # seems to remove use_entities: true ??
        for idx, intent_info in enumerate(data["intents"]):
            for name, intent in intent_info.items():
                if intent.get("use_entities"):
                    data["intents"][idx] = name
```

Am I right in the assumption, that this code from the persist_clean method in domain.py removes the use_entities parameter if it is set to true (and has no other side effects)?

docs/domains.rst
mauricedoepke: Done. I also removed some redundant code. There were 3 places that set a default value for use_entities. The only place in the code where the parameter is actually used is the part that decides whether to featurize it. Thus I removed it from the other 2 places.

rasa/core/domain.py
```diff
@@ -445,10 +440,30 @@ def get_parsing_states(self,
...
-            if should_use_entity:
-                if "entity" in entity:
-                    key = "entity_{0}".format(entity["entity"])
+
+            if "entity" in entity:
+                entity_name = entity["entity"]
+                included_entities = intent_config.get('include_entities', [])
+                excluded_entities = intent_config.get('exclude_entities', [])
+
+                include_default = False if included_entities else True
```
wochinge:
```suggestion
                include_default = not included_entities
```

rasa/core/domain.py
```diff
@@ -229,14 +229,9 @@ def collect_slots(slot_dict):
...
         intent_properties = {}
         for intent in intent_list:
-            if isinstance(intent, dict):
-                for properties in intent.values():
-                    if 'use_entities' not in properties:
-                        properties['use_entities'] = True
-                intent_properties.update(intent)
-            else:
-                intent = {intent: {'use_entities': True}}
-                intent_properties.update(intent)
```
wochinge, line 239: @tmbo This is used by `from_dict`. Do you think we can savely skip providing defaults here and only do it in  `get_parsing_states`?

data/test_domains/default_unfeaturized_entities.yml
```diff
@@ -0,0 +1,22 @@
+intents:
+ - greet: {include_entities: [name]}
+ - default: {exclude_entities : [unrelated_recognized_entity]}
+ - goodbye: {use_entities: false}
```
tmbo: how about `include_entities: false` (or for that matter `exclude_entities: true`), removing the need to have three keys. thoughts?

wochinge: can do :+1:

rasa/core/domain.py
tmbo: mhm, I think we should put the complete properties into `intent_properties` keeping the default values in a single place.

data/test_domains/default_unfeaturized_entities.yml
tmbo: 😆 I am sure we can, the question is more if it is a good idea style wise - would love to hear other peoples opinions (@akelad @erohmensing): Which style of specifying the properties would you prefer?

erohmensing: what about...
 - `{use_entities: None}` instead of `{use_entities: false}` 
 - `{use_entities: List[entities]}` as inclusion
 - `{ignore_entities: List[entities]}` as exclusion?

include/exclude is a little confusing to me because that doesn't say what you're actually doing with them

tmbo: I like that, great suggestion

rasa/core/domain.py
mauricedoepke: The reason I removed this was, that you use this pattern often in the code:
`whatever.get('desired_value', defualt_value_if_not_specified)`

When I did this in `get_parsing_states` this pattern didn't work. Despite not having specified the value in the domain file, the get function didn't return the default value.

Then I found that here in `collect_intent_properties` there was already a default value set, which was different from my desired and as this code was called earlier than mine, my default value got ignored.
This was very confusing and is kind of inconsistent.

I think we should use  `whatever.get('desired_value', defualt_value_if_not_set)` everywhere or set all defaults in one place during loading the config files. Mixing both methods leads to these kinds of bugs described above.

I think setting all defaults in one place, would be the cleaner way. This would also eliminate the possibility of having bugs because of different defaults for the same properties in different places of the code.

wochinge: @mauricedoepke Let's initialize all defaults then in `collect_intent_properties` 👍

data/test_domains/default_unfeaturized_entities.yml
wochinge: @mauricedoepke Then let's do it as @erohmensing suggested it 👍

wochinge:
@mauricedoepke 

> Am I right in the assumption, that this code from the persist_clean method in domain.py removes the use_entities parameter if it is set to true (and has no other side effects)?

Yes

> Do you handle the porting or do I have to do anything to prepare that?

Since it's your time and work, you should also be the author of the lines in the code. But if it's too much effort I can also do it. Should also be just copy past, since their weren't any significant changes to the domain file (just a different location).

rasa/core/domain.py
mauricedoepke: @wochinge

collect_intent_properties is currently a static method which can but does not have to be called to obtain the necessary data for creating a Domain instance. If we initialize all defaults in `collect_intent_properties` and let the class methods rely on that, I suggest we make `collect_intent_properties` a class method and call it in the constructor. The only occasion where `collect_intent_properties` is not directly or indirectly called to create a Domain instance is in one test case which is easy to adapt.

wochinge: It's also used in `_write_domain_to_file` in `core/training/interactive.py`. Putting it in the constructor sounds like a good idea (since we also always call the constructur when we call `collect_intent_properties` ). But I'd actually keep the `classmethod` decorator since it has no `self` references.

wochinge: Sorry for the late feedback @mauricedoepke

erohmensing:
What's the status on this PR @mauricedoepke? Just ran into a forum user who'd definitely benefit from this functionality, I think it's a great addition 😄

mauricedoepke:
@erohmensing 

> What's the status on this PR @mauricedoepke? Just ran into a forum user who'd definitely benefit from this functionality, I think it's a great addition 😄

Sorry for not replying for some time. I had some personal issues which left me with no time.
The feature itself is nearly ready, I still have some refactoring left regarding what I discussed with wochinge. I should be done with it this week.

erohmensing:
That's great to hear @mauricedoepke! We'd love to get this one merged, it's a highly requested feature. Thanks for all the work you've put into it so far!

mauricedoepke:
@wochinge 
I have some trouble with the "tests/core/test_policies.py::TestTwoStageFallbackPolicy::test_persist_and_load" test.

I think the test is broken.

It uses the domain defined on the class level:
```
    @pytest.fixture(scope="class")
    def default_domain(self):
        content = """
        actions:
          - utter_hello

        intents:
          - greet
          - bye
          - affirm
          - deny
        """
        return Domain.from_yaml(content)
```

and then calls the "train_trackers" method with this domain. Which again calls the "training.load_data" method with the given domain and the default stories file. The default stories file, however, contains intents which are not defined in the domain (default and goodbye). Thus my code is failing this test. If I add the missing intents the test works. But as you seem to have added this domain definition and the domain is used in some other tests as well I'd like to know whether you can remember why you used this particular definition instead of the default one and whether it's better to
1) extend the domain definition with these intents
2) use the default domain for this test
3) use a story fitting your domain

wochinge:
@mauricedoepke Thanks for finishing up this pr! I think this test was already fixed in the merged repository `rasa`. Would you mind porting it? I'm also happy to assist you / doing it for you 🚀

mauricedoepke:
@wochinge

Would I just create a branch on the other repo and apply the same changes I made here manually to the other repo? Or what is the proper way to do this?

wochinge:
@mauricedoepke Not sure if you can actually create a branch in the repository. The clean way would probably be to 
1) fork https://github.com/rasahq/rasa 
2) Create a branch with your changes
3) Create a pr against https://github.com/rasahq/rasa (can also be a draft, I'm happy to help you with any issues)

mauricedoepke:
I ported this to the new repo:
https://github.com/RasaHQ/rasa/pull/3655

PR #1689, opened 2019-02-06
added ReminderCancelled event

gkniesel:
**Proposed changes**:
- add ReminderCancelled event
- fixes #1165 

**Status (please check what you already did)**:
- [x] made PR ready for code review
- [ ] added some tests for the functionality
- [ ] updated the documentation
- [ ] updated the changelog

PR #1859, opened 2019-03-18
added ReminderCancelled event

Ghostvv:
**Proposed changes**:
- finish #1689

- [x] updated the changelog


rasa_core/processor.py
```diff
@@ -336,7 +338,23 @@ def should_predict_another_action(action_name, events):
...
+                             tracker.sender_id)
+
+    @staticmethod
+    async def _cancel_reminders(events: List[Event],
+                                tracker: DialogueStateTracker) -> None:
+        # All Reminders with the same name will be cancelled
+        if events is not None:
+            for e in events:
+                if isinstance(e, ReminderCancelled):
+                    name_to_check = e.action_name +\
```
tmbo: should use `(...)` instead of `\`

rasa_core/processor.py
```diff
@@ -336,7 +338,23 @@ def should_predict_another_action(action_name, events):
...
-                        name=str(e.action_name))
+                        name=str(e.action_name) +
+                             ACTION_NAME_SENDER_ID_CONNECTOR_STR +
+                             tracker.sender_id)
+
+    @staticmethod
+    async def _cancel_reminders(events: List[Event],
+                                tracker: DialogueStateTracker) -> None:
+        # All Reminders with the same name will be cancelled
+        if events is not None:
```
tmbo: this can never be `None` according to the type annotation

rasa_core/processor.py
```diff
@@ -336,7 +338,23 @@ def should_predict_another_action(action_name, events):
...
+            for e in events:
+                if isinstance(e, ReminderCancelled):
+                    name_to_check = e.action_name +\
+                                    ACTION_NAME_SENDER_ID_CONNECTOR_STR +\
+                                    tracker.sender_id
+                    for j in (await jobs.scheduler()).get_jobs():
```
tmbo: the wait is ok, but I'd rather do this once before the loop, e.g.
```
scheduler = await jobs.scheduler()
for j in scheduler.get_jobs():
```

tests/test_processor.py
```diff
@@ -119,6 +118,29 @@ def loop():
...
+    t.update(r)
+    t.update(ReminderCancelled("utter_greet"))
+
+    default_processor.tracker_store.save(t)
+    await default_processor.handle_reminder(r, d)
+
+    # retrieve the updated tracker
+    t = default_processor.tracker_store.retrieve(sender_id)
+    # there should be no utter_greet action
+    assert ActionExecuted("utter_greet") not in t.events
```
tmbo: I think we should also check the jobs of the scheduler here making sure it is not scheduled

rasa_core/events/__init__.py
```diff
@@ -567,6 +567,44 @@ def _from_story_string(
                                   parameters.get("timestamp"))]
 
 
+# noinspection PyProtectedMember
+class ReminderCancelled(Event):
+    """ Cancel all jobs with a specific name."""
```
tmbo:
```suggestion
    """Cancel all jobs with a specific name."""
```

rasa_core/processor.py
```diff
@@ -336,7 +338,23 @@ def should_predict_another_action(action_name, events):
...
+                        if j.name == name_to_check:
+                            (await jobs.scheduler()).remove_job(j.id)
```
Ghostvv: @tmbo here `scheduler.remove_job(j.id)` as well?

tests/test_processor.py
```diff
@@ -119,6 +119,53 @@ def loop():
...
+    # check that the jobs were added
+    assert len((await jobs.scheduler()).get_jobs()) == 2
+
+    for t in trackers:
+        await default_processor._cancel_reminders(t.events, t)
+    # check that only one job was removed
+    assert len((await jobs.scheduler()).get_jobs()) == 1
+
+    # # execute the second reminder
+    await default_processor.handle_reminder(reminders[1], dispatchers[1])
```
Ghostvv: @tmbo I don't like the line here

PR #1861, opened 2019-03-18
Intent evaluation endpoint

### LINKED ISSUE #1860: Intent evaluation endpoint implementation

ricwo:
**Proposed changes**:
- fixes #1860 

- [x] updated the changelog


docs/_static/spec/server.yml
```diff
@@ -655,6 +655,77 @@ paths:
...
+      tags:
+      - Model
+      summary: Perform an intent evaluation on a Rasa Stack model
+      description: >-
+        Evaluates NLU intents on a Rasa Stack model. A .tar.gz file has to be
+        provided which must contain the NLU evaluation data and a Rasa Stack
+        model. Example contents of the zipped directory (the NLU evaluation
+        data is contained in data.json):
+
+        evaluation.tar.gz
```
tmbo: I am not convinced this is the best structure, maybe multi-part-form submission is better?

I actually think we shouldn't have these endpoints in the first place and rather use some sort of task management / worker system to submit trainings / evaluations. but I guess for the moment this is the easiest solution.

rasa_core/server.py
```diff
@@ -534,6 +551,38 @@ def create_app(agent=None,
...
                                 "".format(e))
 
+    @app.post("/intentEvaluation")
+    @requires_auth(app, auth_token)
+    async def evaluate_intents(request: Request):
+        """Evaluate intents against a Rasa NLU model."""
+
+        tmpdir = tempfile.mkdtemp()
+
+        zipped_model_path = os.path.join(tmpdir, 'model.tar.gz')
```
tmbo: this sounds like something we are also doing in the training endpoint, probably makes sense to move the "take body and write to file" into a function

rasa_core/server.py
```diff
@@ -534,6 +551,38 @@ def create_app(agent=None,
...
+        with open(zipped_model_path, 'w+b') as f:
+            f.write(request.body)
+
+        unzipped_path = unpack_model(zipped_model_path, tmpdir)
```
tmbo: we should move this logic out of the endpoint into a separate function (so that we can test separately without running the server)

rasa_core/server.py
```diff
@@ -166,6 +169,20 @@ def event_verbosity_parameter(request, default_verbosity):
                             {"parameter": "include_events", "in": "query"})
 
 
+async def find_nlu_files_in_path(path: Text):
```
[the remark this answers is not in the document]
wochinge: https://github.com/RasaHQ/rasa_core/blob/f34fcbc0c2d2264d9448fc40207e29d21ea26280/rasa/data.py#L26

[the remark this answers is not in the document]
ricwo: This function won't work as it collects NLU files recursively. The NLU model dir always contains `training_data.json` which we don't need in this case

rasa_core/server.py
ricwo: hmm, we don't dump the whole body in the training endpoint, but just the json payload. will create a function though

ricwo:
@tmbo Looks like it's a `LocalPath` in py3.5

ricwo:
https://travis-ci.com/RasaHQ/rasa_core/jobs/186261952#L1899

PR #1873, opened 2019-03-20
Logging fallback policy

### LINKED ISSUE #1847: add logging for low core confidence for FallbackPolicy
logger debug should definitely tell you when Fallback was invoked because of low `core_threshold`, so that people know it was the action prediction and not the intent classification that messed up. 

Maybe a good addition if it doesn't require adding too much: logging something like <next_action> predicted by <(machine learning) policy> with confidence <confidence> is below core threshold <core_threshold>

tabergma:
**Proposed changes**:
- Always evaluate FallbackPolicy last
- If FallbackPolicy was chosen in the end and the confidence is equal to the core threshold, we know that all other action prediction confidence scores were lower than the core threshold

closes #1847 

- [ ] updated the changelog

### DIFF
--- a/rasa_core/policies/ensemble.py
+++ b/rasa_core/policies/ensemble.py
@@ -354,14 +354,16 @@ class SimplePolicyEnsemble(PolicyEnsemble):
                 best_policy_index = i
 
         if logging.getLogger().level == logging.DEBUG:
+            # Logger should inform you when fallback was invoked because of low core_threshold,
+            # so people know it was the action prediction and not the intent classification that
+            # messed up.
             fallback_idx_policy = [(i, p) for i, p in enumerate(self.policies)
                                     if isinstance(p, FallbackPolicy)]
-
             if fallback_idx_policy:
                 fallback_idx, fallback_policy = fallback_idx_policy[0]
                 if fallback_idx == best_policy_index and fallback_policy.core_threshold_set:
-                     logger.debug("fallback policy was invoked because of core_threshold")
-
+                    logger.debug("Fallback policy was invoked because of core_threshold ({})".format(
+                        fallback_policy.core_threshold))
 
         if (result.index(max_confidence) ==
                 domain.index_for_action(ACTION_LISTEN_NAME) and
--- a/rasa_core/policies/ensemble.py
+++ b/rasa_core/policies/ensemble.py
@@ -362,7 +362,7 @@ class SimplePolicyEnsemble(PolicyEnsemble):
             if fallback_idx_policy:
                 fallback_idx, fallback_policy = fallback_idx_policy[0]
                 if fallback_idx == best_policy_index and \
-                        fallback_policy.core_threshold_set:
+                        fallback_policy.should_core_fallback:
                     logger.debug("Fallback policy was invoked because of "
                                  "core_threshold ({})"
                                  .format(fallback_policy.core_threshold))
--- a/rasa_core/policies/ensemble.py
+++ b/rasa_core/policies/ensemble.py
@@ -331,10 +331,24 @@ class SimplePolicyEnsemble(PolicyEnsemble):
                                         tracker: DialogueStateTracker,
                                         domain: Domain
                                         ) -> Tuple[List[float], Text]:
+
+        fallback_idx_policy = [(i, p) for i, p in enumerate(self.policies)
+                               if isinstance(p, FallbackPolicy)]
+
+        fallback_idx, fallback_policy = -1, None
+        if fallback_idx_policy:
+            fallback_idx, fallback_policy = fallback_idx_policy[0]
+            # evaluate the fallback policy last, so that we are able to
+            # determine that there is no other policy with higher action
+            # prediction confidence than the core threshold
+            self.policies.append(self.policies.pop(fallback_idx))
+            fallback_idx = self.policies.index(fallback_policy)
+
         result = None
         max_confidence = -1
-        best_policy_name = None
+        best_policy = None
         best_policy_priority = -1
+        best_policy_name = None
 
         for i, p in enumerate(self.policies):
             probabilities = p.predict_action_probabilities(tracker, domain)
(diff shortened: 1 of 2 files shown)


rasa_core/policies/fallback.py
```diff
@@ -87,7 +88,7 @@ def predict_action_probabilities(self,
         The fallback action is predicted if the NLU confidence is low
         or no other policy has a high-confidence prediction.
         """
-
+        self.core_threshold_set = False
```
erohmensing: i like the implementation of the flag, but think the `core_threshold_set` name could be confusing (`core_threshold_set` makes me think more that the user overwrote the default threshold in their policy config). I think the action is something more similar to `should_nlu_fallback`, so maybe `should_core_fallback`?

rasa_core/policies/fallback.py
```diff
@@ -47,6 +47,7 @@ def __init__(self,
         self.nlu_threshold = nlu_threshold
         self.core_threshold = core_threshold
         self.fallback_action_name = fallback_action_name
+        self.core_threshold_set = False
```
erohmensing: Do we need this in the initialization here if it's only used in the `predict_action_probabilities`?

rasa_core/policies/ensemble.py
```diff
@@ -350,6 +351,21 @@ def probabilities_using_best_policy(self,
                 result = probabilities
                 best_policy_name = 'policy_{}_{}'.format(i, type(p).__name__)
                 best_policy_priority = p.priority
+                best_policy_index = i
+
+        if logger.level == logging.DEBUG:
```
erohmensing: Not sure about our general coding practices here (perhaps some input from @tmbo would help) -- it makes sense to only go through this logic if the logger is on debug, but I don't see anywhere else in the code where take the logger level into account

rasa_core/policies/ensemble.py
```diff
@@ -335,6 +335,7 @@ def probabilities_using_best_policy(self,
         max_confidence = -1
         best_policy_name = None
         best_policy_priority = -1
+        best_policy_index = -1
```
erohmensing: maybe we can clean this up together -- I just added the `best_policy_priority` when I added that feature, but perhaps should have questioned my adding of the variable instead of changing the one already there. `best_policy_name`, `best_policy_priority` and `best_policy_index` are all coming from the same thing -- could we perhaps just keep track of the `best_policy` and then get the name and priority from that? (the index is already in the name)

rasa_core/policies/ensemble.py
```diff
@@ -350,6 +351,21 @@ def probabilities_using_best_policy(self,
                 result = probabilities
                 best_policy_name = 'policy_{}_{}'.format(i, type(p).__name__)
                 best_policy_priority = p.priority
+                best_policy_index = i
+
+        if logger.level == logging.DEBUG:
+            # When fallback was invoked because of low core_threshold,
+            # log the information so that people know it was the action
+            # prediction and not the intent classification that messed up.
```
erohmensing: Maybe state this a little more elegantly than I did 😅

rasa_core/policies/ensemble.py
```diff
@@ -350,6 +351,21 @@ def probabilities_using_best_policy(self,
...
+            fallback_idx_policy = [(i, p) for i, p in enumerate(self.policies)
+                                   if isinstance(p, FallbackPolicy)]
+            if fallback_idx_policy:
+                fallback_idx, fallback_policy = fallback_idx_policy[0]
+                if fallback_idx == best_policy_index and \
+                        fallback_policy.core_threshold_set:
+                    logger.debug("Fallback policy was invoked because of "
+                                 "core_threshold ({})"
```
erohmensing: Perhaps something like `"Highest action prediction confidence is lower than core threshold {}.".format(core_threshold)`
(logger will already tell you that fallback policy was used to predict the next action, I think the goal here is more to clarify that there wasn't a confident enough action to get above the threshold)

rasa_core/policies/fallback.py
tabergma: As we are accessing the variable outside the `FallbackPolicy` class in the ensemble, we need the initialization.

rasa_core/policies/fallback.py
tmbo: This flag wont work: The policy (as part of the model) is shared by all the conversations. Because we are using async, there should be no state within that class that is specific to a conversation (as the server might receive a message while we are still handling one, due to asyncio we might jump into handling that other message, even though the first one is not finished yet)

erohmensing: > there should be no state within that class that is specific to a conversation

@tmbo Good to know -- I'd love a clarification sometime about how async changes things, and what mistakes like this to avoid as a result

erohmensing:
@tmbo this will have to be rebased into the `rasa_nlu` repo before merge right?

tmbo:
yes

rasa_core/policies/fallback.py
tabergma: Ok. In that case we might go with the idea of @erohmensing: evaluate the fallback policy always at last and log the message if the fallback policy was chosen and the obtained confidence is equal to the core_threshold.

rasa_core/policies/ensemble.py
```diff
@@ -331,10 +331,24 @@ def probabilities_using_best_policy(self,
...
+                               if isinstance(p, FallbackPolicy)]
+
+        fallback_idx, fallback_policy = -1, None
+        if fallback_idx_policy:
+            fallback_idx, fallback_policy = fallback_idx_policy[0]
+            # evaluate the fallback policy last, so that we are able to
+            # determine that there is no other policy with higher action
+            # prediction confidence than the core threshold
+            self.policies.append(self.policies.pop(fallback_idx))
+            fallback_idx = self.policies.index(fallback_policy)
```
erohmensing: I worry that changing the index of the fallback policy might mess things up when at the end of this method we return the name of the best policy (which includes the index)

tabergma:
Is it worth changing all of this, just to log a simple message? It seems to be quite complicated just for a single log line.

Alternative idea:
Log a message in the `FallbackPolicy` itself when the `core_threshold` is used as confidence. In case the `FallbackPolicy` was chosen in the end, the user knows from the log message before, that there was no other action policy that had a higher confidence than the `core_threshold`. If the `FallbackPolicy` was not chosen, the user can ignore the message from the fallback policy.

What do you think?

tmbo:
Sounds good, simple and suits the need ⭐️

erohmensing:
+1. I think that sounds like a much better idea. Should we just close this one and make those changes in NLU once core is in there so we don't have to port?

tabergma:
Yes. Would close this PR and open the same issue on NLU again. Thanks for your input.

PR #1879, opened 2019-03-25
fix evaluation with forms

Ghostvv:
**Proposed changes**:
- emulate form rejection

- [ ] updated the changelog

### DIFF
--- a/examples/formbot/Makefile
+++ b/examples/formbot/Makefile
@@ -12,7 +12,7 @@ help:
 
 
 train-core:
-	python -m rasa_core.train -s data/stories.md -d domain.yml -o models/dialogue --debug
+	python3 -m rasa_core.train -s data/stories.md -d domain.yml -o models/dialogue --debug
 
 run-core:
 	python -m rasa_core.run --core models/dialogue --nlu models/nlu/current --debug --endpoints endpoints.yml
@@ -32,3 +32,6 @@ visualize:
 
 train-nlu:
 	python -m rasa_nlu.train -c nlu_tensorflow.yml --fixed_model_name current --data data/nlu_data.md -o models --project nlu --verbose
+
+evaluate-core:
+	python3 -m rasa_core.evaluate --core models/dialogue -s data/stories.md --fail_on_prediction_errors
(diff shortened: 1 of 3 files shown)


Ghostvv:
@akelad could you please take a look at the emulation of form rejection logic? What do you think about assuming that form was rejected if FormPolicy memorized such unhappy state in training stories?

examples/formbot/Makefile
```diff
@@ -12,7 +12,7 @@ help:
 
 
 train-core:
-	python -m rasa_core.train -s data/stories.md -d domain.yml -o models/dialogue --debug
+	python3 -m rasa_core.train -s data/stories.md -d domain.yml -o models/dialogue --debug
```
akelad: can you get rid of the python3 stuff please

rasa_core/policies/form_policy.py
```diff
@@ -78,6 +78,28 @@ def recall(self,
...
+        states = tracker_as_states[0]
+
+        memorized_form = self.recall(states, tracker, domain)
+
+        state_is_unhappy = (
+            memorized_form is not None and
+            memorized_form == tracker.active_form.get('name')
+        )
+
+        if state_is_unhappy:
```
akelad: doesn't it make sense to leave the debug statement where it was initally? to avoid doing the if twice?

Ghostvv: it makes, but then in the second place I'll need to recalculate states

Ghostvv: or pass the states instead of tracker, which makes using this function less convenient

akelad: ok fair, in that case feel free to merge. Have you opened a PR against NLU master as well yet?

erohmensing: @akelad that would be [this one](https://github.com/RasaHQ/rasa_nlu/pull/3140#issuecomment-477933594) i believe

akelad: oh man 🙈 so hard to keep track of everything haha. @Ghostvv can you make sure the changes are identical on both because they don't seem to be

PR #1880, opened 2019-03-25
fixing attachment in utter_response

mjspeck:
**Proposed changes**:
- Doesn't seem like this made any sense. An image associated with text was being added as the attachment.

- [ ] updated the changelog


akelad:
@mjspeck could you open a PR against the NLU repo instead please? we're no longer updating master of this repository

PR #1883, opened 2019-04-02
Extract message intent instead of setting the node with a dict when using visualization

MetcalfeTom:
Fixes https://github.com/RasaHQ/rasa_nlu/issues/2995

Under certain circumstances we were passing the message parse data to the node label during interactive learning visualization, which (since this is a dictionary) was causing fatal errors.

**Proposed changes**:
- Use message intent name instead of the parse data as the graph node label

- [x] updated the changelog

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,14 @@ Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to `Semantic Versioning`_ starting with version 0.2.0.
 
+[Unreleased 0.13.a8]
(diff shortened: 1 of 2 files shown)


MetcalfeTom:
@paulaWesselmann

Just a convenient way of passing both the text and the intent; keeping the text makes the graph readable and keeping the intent allows us to compare equivalent nodes.  If the same intent has appeared twice, we overwrite the node label with the subsequent message text.

paulaWesselmann:
Approved because it error proofs this from happening. But I'm still confused on how it comes to this sequence of events that it is able to happen at all.

PR #1894, opened 2019-04-26
fix bug and outdated changelog

erohmensing:
**Proposed changes**:
- fix mapping policy bug
- fix https://github.com/RasaHQ/rasa/issues/3328

- [x] updated the changelog

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,14 @@ Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to `Semantic Versioning`_ starting with version 0.2.0.
 
+[0.14.1] - unreleased
+^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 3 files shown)


erohmensing:
Yes figured out why it worked with the one of that guy's actions, because the bug shows up in action prediction after the mapped action when mappingpolicy predicts action listen -- his one action had a follow up action so didn't go into the logic with the bug in it. 

As for the tests, that test isn't an actual mappingpolicy test, just one to make sure that the domain is read correctly. The MappingPolicy has no real implementation tests added on top of the regular policy loading, persisting etc. I was looking into adding tests but figured i'd get the fix out first so we could update it since its pretty broken.

We've made a decent amount of fixes to CLI bugs on master that could be pulled into 0.14.1 but they're not all merged (one introduced a bug that's fixed on another branch that's not yet merged)

PR #1903, opened 2019-05-17
Update Rasa Core readme

alexweidauer:
**Proposed changes**:
- ...

**Status (please check what you already did)**:
- [ ] made PR ready for code review
- [ ] added some tests for the functionality
- [ ] updated the documentation
- [ ] updated the changelog

PR #1904, opened 2019-05-20
update timeout

erohmensing:
I know we updated this timeout on `rasa` (though it seems we might have gotten rid of it temporarily)? -- [This PR](https://github.com/RasaHQ/rasa_core/pull/1903) keeps failing


erohmensing:
or... not. should we just remove the timeout here too?

